PR #11, opened 2023-03-17
Environment and development usability improvements

goldmann:
This PR includes a long list of fixes and usability improvements. Main focus was on modularity of the kustomize scripts, but this triggered a unified way of preparing development environment as well. Now we have two overlays that can be used for developemnt:

- `local`
- `development`

The `development` one should be used on daily basis when we are adding new features and we want to see them live. It DOES NOT deploy the service inside of the Minikube environment. Instead, it lets the developer run it outside (in IDE?). For development -- look at the `hack/build*` scripts which make building images a breeze. Especially the `build-images-minikube.sh` which makes these images available inside the Minikube environment instantly.

The `local` one deploys everything into local Minikube cluster.

The `hack` directory contains many helpful scripts.

Besides this -- tons of little tweaks here and there.


vibe13 (approved):
Massive work, looks great!!!

PR #12, opened 2023-03-17
Refactor model fixes

vibe13:
- Added generator and processor enums to the Sbom model
- Create qualifiers for the different SbomGenerator implementations
- Renamed transformer to processor; create qualifiers for the different SbomProcessor implementations; removed SbomManipulator
- Removed enrichment task; fixed payload to create SBOMs after generation
- Update of the service and creation of save and enrichment methods
- Update REST APIs with multiple endpoints to generate using Domino and CycloneDX, and to process using various processor (there is only one now)


k8s/base/tekton/resources/task-base-enrichment.yaml
goldmann, line 1: Should be removed from `k8s/base/tekton/kustomization.yaml` as well.

src/main/java/org/redhat/sbomer/rest/v1alpha1/SBOMResource.java
```diff
@@ -116,50 +99,77 @@ public Page<Sbom> list(@Valid @BeanParam PaginationParameters paginationParams)
...
-            description = "SBOM enrichment for a particular PNC build Id. The only mode currently available is `ENRICHED_v1_0`")
+            summary = "Generate a base SBOM based on the PNC build using CycloneDX plugin.",
+            description = "SBOM base generation for a particular PNC build Id offloaded to the service, using CycloneDX plugin.")
     @Parameter(name = "buildId", description = "PNC build identifier", example = "ARYT3LBXDVYAC")
-    @Parameter(
-            name = "mode",
-            description = "Enrichment generation mode. Available modes: `ENRICHED_v1_0`",
-            example = "ENRICHED_v1_0")
-    @Path("/enrich/{buildId}")
+    @Path("/generate/cyclonedx/{buildId}")
```
goldmann, line 142: I wonder if we are interested in multiple endpoints like this or maybe it would be better to expose one: `/generate`. It could take an optional `@PathParam`. By default it would be `CYCLONEDX`, but if needed, it could be overriden. This way we would make it more flexible to add/remove generators. The other benefit is that we would actually point users which one is preferred (default). Not sure if they really care about what is the implementation.

src/main/java/org/redhat/sbomer/rest/v1alpha1/SBOMResource.java
```diff
@@ -116,50 +99,77 @@ public Page<Sbom> list(@Valid @BeanParam PaginationParameters paginationParams)
...
+
+        return sbomService.generateSbomFromPncBuild(id, Generators.CYCLONEDX);
+    }
 
+    @POST
+    @Operation(
+            summary = "Save the base SBOM and run and enrichment with PncToSbomProperties processor.",
+            description = "Save the base SBOM and run and enrichment with PncToSbomProperties processor.")
+    @Parameter(name = "sbom", description = "The SBOM to save")
+    @Path("/process/sbomproperties")
```
goldmann, line 157: Similar comment to previous one. If we would make it just `/process`  and move to params we could simplify the API and make it more less coupled.

src/main/java/org/redhat/sbomer/rest/v1alpha1/SBOMResource.java
```diff
@@ -116,51 +101,93 @@ public Page<Sbom> list(@Valid @BeanParam PaginationParameters paginationParams)
...
+            description = "Generator to use to generate the SBOM. If not specified, CycloneDX will be used. Options are `DOMINO`, `CYCLONEDX`",
+            example = "CYCLONEDX")
+    @Path("/generate/{buildId}")
     @APIResponses({ @APIResponse(
             responseCode = "201",
-            description = "Executes the enrichment of an existing SBOM for a particular PNC buildId.",
+            description = "Schedules generation of a SBOM for a particular PNC buildId. This is an asynchronous call. It does execute the generation behind the scenes.",
             content = @Content(mediaType = MediaType.APPLICATION_JSON)) })
-    public Response runEnrichmentOfBaseSbom(@PathParam("buildId") String buildId, @QueryParam("mode") String mode)
+    public Response generateFromBuildWithDomino(
```
goldmann, line 138: Nitpicking, but I think this could be renamed to just `generate` or similar.

src/main/java/org/redhat/sbomer/rest/v1alpha1/SBOMResource.java
```diff
@@ -116,51 +101,93 @@ public Page<Sbom> list(@Valid @BeanParam PaginationParameters paginationParams)
...
+
+    @POST
+    @Operation(
+            summary = "Save the base SBOM and run and enrichment.",
+            description = "Save the base SBOM and run and enrichment..")
+    @Parameter(name = "sbom", description = "The SBOM to save")
+    @Parameter(
+            name = "processor",
+            description = "Processor to use to enrich the SBOM. If not specified, SBOM_PROPERTIES will be used. Options are `SBOM_PROPERTIES`",
+            example = "CYCLONEDX")
```
goldmann, line 164: I think example should state: `SBOM_PROPERTIES`.

goldmann (approved):

vibe13:
@goldmann thanks for you review and hawk eyes!

PR #17, opened 2023-03-21
SBOMer new Tekton Task API

goldmann:
Initial work on establishing API for all our generator Tekton Tasks.

Changes:

- We are now running `TaskRun`s instead of `PipelineRun`s
- Simplified fetching of source (drop of the `git-clone` Task)
- Task API: just two parameters: `build-id` and `config`. The `config` parameter is currently a JSON, but read the comment there please.
- Dropped usage of PVCs
- Drop the use of env initialization
- Added tests


k8s/base/tekton/resources/task-generate-cyclonedx.yaml
```diff
@@ -22,47 +22,114 @@ metadata:
...
+    - name: checkout
       image: sbomer-generator
       imagePullPolicy: IfNotPresent
       script: |
         #!/usr/bin/env bash
 
+        set -e
+
+        # Fetch build information
+        curl https://orch.psi.redhat.com/pnc-rest/v2/builds/$(params.build-id) | jq > $(workspaces.data.path)/build.json
```
vibe13, line 76: should we expose the orch URL inside the config?

vibe13 (commented):
Looks good, I like the removal of few steps in the pipeline, and the config object to expose settings will probably be much useful when we will need to add more configs! Probably the orch URL could be another one to put already?

k8s/base/tekton/resources/task-generate-cyclonedx.yaml
goldmann: This is very temporary until we convert this into CLI, then we will be running a Quarkus CLI there, with full support for configuration.

goldmann: > Probably the orch URL could be another one to put already?

I don't see this URL being something that users should control. We will be, since we will be pointing to the proxy. See the point about CLI.

goldmann:
> Probably the orch URL could be another one to put already?

I don't see this URL being something that users should control. We will be, since we will be pointing to the proxy. See the point about CLI.

k8s/base/tekton/resources/task-generate-cyclonedx.yaml
vibe13: right those should be user exposed configs. ok all good then!

vibe13 (approved):

PR #16, opened 2023-03-21
Implement new Sbom enricher using pedigree fields

vibe13:
- Add new processor to put additional PNC properties into SBOM pedigree section
- Add SbomType to the model
- Pass environment attributes to task-build-env task to avoid another fetching from PNC; propagate the processor through generate and enrich endpoints inside the pipeline
- Add the type to SBOM
- Add type to the new enriched Sbom

PR #18, opened 2023-03-21
New enrichment processor based on Sbom pedigree (stripped version of https://github.com/project-ncl/sbomer/pull/16)

vibe13:
- Add the type to SBOM
- Implement new enrichment processor based on Sbom pedigree, and make it the default processor

### DIFF
--- a/src/main/java/org/redhat/sbomer/processor/PncArtifactsToSbomPedigreeProcessor.java
+++ b/src/main/java/org/redhat/sbomer/processor/PncArtifactsToSbomPedigreeProcessor.java
@@ -1,0 +1,103 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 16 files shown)


src/main/java/org/redhat/sbomer/processor/PncArtifactsToSbomPedigreeProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+@PncToSbomPedigree
+@ApplicationScoped
+public class PncArtifactsToSbomPedigreeProcessor implements SbomProcessor {
+
+    @Inject
+    SBOMService sbomService;
+
+    @Override
+    public Bom process(Bom originalBom) {
+        log.info("Adding PNC cached build info to the SBOM properties");
```
goldmann, line 55: We probably should add more info to our logs, especially what SBOM we are talking here about.

goldmann (approved):
Good stuff! The only thing that it would be good to add is documentation to the endpoint in usage docs. That document wasn't updated after last refactoring (but it should!). So, my proposal here is to basically merge this one and update the docs as part of https://github.com/project-ncl/sbomer/issues/19.

The other comment above is a nit. Decide if you want to add more info, if not, just go ahead and merge it!

vibe13:
Thanks Marek! I will update the docs as you said and add more useful logs in a future PR :-) thanks!

PR #21, opened 2023-03-22
ci(hack): fix minikube image generation

goldmann:
It looks that after last refactoring of scripts I introduced a bug in Minikube image generation.


goldmann:
Merging this one myself, because it messes my workflow :)

PR #30, opened 2023-03-24
Introduce multi-module project

goldmann:
* Introduction of a multi-module project, currently with 3 modules: **core** (things that can/should be shared), **cli** (command line executor) and **service** (REST API and main service). Some code was moved already to appropriate modules, but more refactoring will follow!
* Package rename for consistency to `org.jboss.sbomer` + module suffix.
* Introduction of a CLI based on Quarkus and PicoCLI. Currently it is pretty much a skeleton that does nothing, but it will be populated with content over this week.
* Refactoring of CDI bean selectors, dropped usage of `@CycloneDX` and `@Domino` annotations and currently a single `@Generator` is used: `@Generator(CYCLONEDX)` and `@Generator(DOMINO)`. Similarly, for processors -- dropped `@PncToSbomProperties` and `@PncToSbomPedigree` in favor of `@Processor(PEDIGREE)` and `@Processor(PROPERTIES)`
* Moved container image sources to `images/` dir in root.


service/src/main/java/org/jboss/sbomer/processor/PncArtifactsToSbomPedigreeProcessor.java
```diff
@@ -15,50 +15,46 @@
...
-        log.info("Applying SBOM_PEDIGREE processing to the SBOM: {}", originalSbom);
-
-        Bom originalBom = originalSbom.getCycloneDxBom();
-        if (originalBom == null) {
-            throw new ValidationException("Could not convert initial SBOM of build: " + originalSbom.getBuildId());
-        }
+    public Bom process(Bom originalBom) {
+        log.info(
+                "Applying SBOM_PEDIGREE processing to the SBOM: {}",
+                originalBom.getMetadata().getComponent().getPurl());
```
vibe13, line 57: My only concern here is that `metadata` is not a required field according to 1.4 spec: https://cyclonedx.org/docs/1.4/json/#metadata_component, so we are not sure this will always be populated (we haven't looked into other build type generators). So although when I changed the signature some days ago from Bom to Sbom I did not like it 100%, not sure if we should change it back _now_

goldmann: I think there was something as well behind the change -- inc case it will be moved to core/cli -- we would not depend at all on the entity.

vibe13: yes the entity is not used beside the Bom so this is the same reason why I did not like it 100%. ok so maybe we can verify if the metadata is not null and in case log other fields like `serialNumber` (but still not mandatory in the spec), or we could pick the first element in `components` (still not mandatory and not sure about the ordering priority).... but anyway this is a log and can be improved if needed

PR #32, opened 2023-03-28
CLI improvements

goldmann:
Ok, so this one is a big one. Wanted to start small, by just moving some bits from service into CLI, but it triggered other things which triggered other things... So, yes, I apologize for this crappy PR.

What did change? A lot!

* First of all, all generation and all processing is now done via the CLI. This made the service more "dumb". I think there are still areas where it can be improved, but I would like to drop this huge PR first.
  * To be checked on the property processor, setting the `SBOM_RED_HAT_BUILD_SYSTEM`
  * The service does not call PNC now, it's the CLI
* Improvement error handling, both on the REST API and in the CLI as well.
* Some basic tests for the CLI were added, but this needs to be extended.
* Introduced wiemock to test CLI interaction with the service
* Rename of the "pedigree" processor into "default"
* After saving a base SBOM, we trigger automatically another task to run the default processor
* Changes in annotations and enums. Now we use a bit more clean approach to select processor/generator. See `Processor` and `Generator` annotations.
* Additional arguments to tools are not yet covered


goldmann:
Added some notes above. What I still need to do is to update tasks to run the CLI properly. I'll do it tomorrow.

cli/src/main/java/org/jboss/sbomer/cli/commands/processor/DefaultProcessCommand.java
```diff
@@ -0,0 +1,78 @@
...
+        SbomUtils
+                .addExternalReference(component, ExternalReference.Type.DISTRIBUTION, Constants.MRRC_URL, DISTRIBUTION);
+        SbomUtils.addExternalReference(
+                component,
+                ExternalReference.Type.BUILD_META,
+                artifact.getBuild().getEnvironment().getSystemImageRepositoryUrl() + "/"
+                        + artifact.getBuild().getEnvironment().getSystemImageId(),
+                SBOM_RED_HAT_ENVIRONMENT_IMAGE);
+        if (!SbomUtils.hasExternalReference(component, ExternalReference.Type.VCS)
+                && !Strings.isEmpty(artifact.getBuild().getScmRepository().getExternalUrl())) {
```
vibe13, line 56: this was probably not rebased on latest commits.

core/src/main/java/org/jboss/sbomer/core/utils/SbomUtils.java
```diff
@@ -88,8 +98,17 @@ public static Optional<Property> findPropertyWithNameInComponent(String property
...
+        List<ExternalReference> filteredExternalReferences = Optional.ofNullable(c.getExternalReferences())
+                .map(Collection::stream)
+                .orElseGet(Stream::empty)
+                .filter(ref -> ref.getType().equals(type))
+                .toList();
+
+        return filteredExternalReferences;
     }
 
     public static void addExternalReference(Component c, ExternalReference.Type type, String url, String comment) {
```
vibe13, line 114: this was probably not rebased on latest commits.

vibe13 (approved):
Hey Marek!!! My eyes are bleeding LOL but this is great stuff!!! Thanks a lot for working on these, code is so much cleaner now. I believe you need to rebase on latest commits as I made a few small fixes around. Other than that, +100!!!

core/src/main/java/org/jboss/sbomer/core/utils/SbomUtils.java
goldmann: I just checked this one, and it looks exactly like it is in latest `main`. Am I missing something?

goldmann:
Thanks a lot for the review!

I really feel guilty and I promise this is the last big one :)

cli/src/main/java/org/jboss/sbomer/cli/commands/processor/DefaultProcessCommand.java
vibe13: see https://github.com/project-ncl/sbomer/blob/main/service/src/main/java/org/jboss/sbomer/processor/PncArtifactsToSbomPedigreeProcessor.java#L86, there is no more the second check

goldmann: Aaah, this thing. Let me fix that while doing the rebase!

core/src/main/java/org/jboss/sbomer/core/utils/SbomUtils.java
vibe13: sorry my bad, looks ok. I was confused by previous comment.

goldmann:
@vibe13 So, I've added all changes that make it possible to run in local mode. Previously I was testing only development mode. There are still many things to improve, but I would like to tackle these in subsequent PRs.

goldmann:
I'm going to get ahead and merge it!

PR #37, opened 2023-04-12
Implement resource statuses

goldmann:
Main change in this PR is that **resources have now statuses**.

## Generating

Now, when a generation is requested, an `Sbom` object is created and proper status set. Below is an example response returned to the requester when :

```json
HTTP/1.1 202 Accepted
Content-Type: application/json;charset=UTF-8
content-length: 230

{
    "buildId": "ARYT3LBXDVYAC",
    "generationTime": "2023-04-17T09:10:52.753561090Z",
    "generator": "CYCLONEDX",
    "id": 435728847752626176,
    "parentSbom": null,
    "processor": null,
    "rootPurl": null,
    "sbom": null,
    "status": "GENERATING",
    "type": "BUILD_TIME"
}
```

Things to note:

1. `sbom` is null -- obviosly, we haven't generated it yet.
2. `status` is `GENERATING`
3. HTTP code is `202`, accepted.

This lets the client understand the status and at the same time have a pointer to the resource: id: `435728847752626176`, so the status (and resource) can be queried by using this path: `/api/v1alpha1/sboms/435728847752626176`.

Underneath, when the generation process finishes, we are just updating the `sbom` field in the resource. Additionally the `rootPurl` is updated as well.

## Processing

Another change is that currently we are NOT triggering automatically the processing of the generated SBOM.

Similarly to generation, processing returns a similar response:

```json
HTTP/1.1 202 Accepted
Content-Type: application/json;charset=UTF-8
transfer-encoding: chunked

{
    "buildId": "ARYT3LBXDVYAC",
    "generationTime": "2023-04-17T09:16:38.909539596Z",
    "generator": "CYCLONEDX",
    "id": 435730299631927296,
    "parentSbom": { ... }
    "processor": "DEFAULT",
    "rootPurl": null,
    "sbom": null,
    "status": "PROCESSING",
    "type": "BUILD_TIME"
}
```

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/commands/generator/AbstractMavenBaseGenerateCommand.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/commands/generator/AbstractMavenBaseGenerateCommand.java
@@ -39,7 +39,6 @@ import org.jboss.sbomer.cli.client.SBOMerClient;
 import org.jboss.sbomer.cli.model.Sbom;
 import org.jboss.sbomer.cli.service.PNCService;
 import org.jboss.sbomer.core.enums.GeneratorImplementation;
-import org.jboss.sbomer.core.enums.SbomType;
 import org.jboss.sbomer.core.errors.ApplicationException;
 import org.jboss.sbomer.core.utils.SbomUtils;
 
@@ -71,11 +70,14 @@ public abstract class AbstractMavenBaseGenerateCommand implements Callable<Integ
 
     @Override
     public Integer call() throws Exception {
+        // First, fetch the SBOM metadata
+        Sbom sbom = sbomerClient.getById(parent.getParent().getSbomId());
+
         // Fetch build information
-        Build build = pncService.getBuild(parent.getParent().getBuildId());
+        Build build = pncService.getBuild(sbom.getBuildId());
 
         if (build == null) {
-            log.error("Could not fetch the PNC build with id '{}'", parent.getParent().getBuildId());
+            log.error("Could not fetch the PNC build with id '{}'", parent.getParent().getSbomId());
             return CommandLine.ExitCode.SOFTWARE;
         }
 
@@ -85,25 +87,9 @@ public abstract class AbstractMavenBaseGenerateCommand implements Callable<Integ
         // Generate the SBOM
         Path bomPath = generate();
 
-        doUpload(bomPath);
-
-        return CommandLine.ExitCode.OK;
-    }
-
-    protected abstract GeneratorImplementation getGeneratorType();
-
-    protected abstract Path generate();
-
-    /**
-     * Responsible for uploading the generated SBOM to the service using a REST endpoint.
-     *
-     * @param bomPath The path to the SBOM file
-     * @throws ParametrizedException In case processing or upload fails.
-     */
-    protected void doUpload(Path bomPath) {
         log.info(
                 "Preparing to upload SBOM for build '{}' from path '{}' to the service...",
-                parent.getParent().getBuildId(),
+                parent.getParent().getSbomId(),
                 bomPath);
 
         log.debug("Reading generated SBOM from '{}' path", bomPath);
@@ -115,23 +101,18 @@ public abstract class AbstractMavenBaseGenerateCommand implements Callable<Integ
             throw new ApplicationException("Could parse the generated SBOM from '{}' path", bomPath.toString());
         }
 
-        log.debug("Creating SBOM object out of the generated SBOM...");
+        log.info("Uploading generated CycloneDX BOM...");
 
-        // Create the Sbom object
-        Sbom sbom = new Sbom();
-        sbom.setBuildId(parent.getParent().getBuildId());
-        sbom.setGenerator(getGeneratorType());
-        sbom.setSbom(SbomUtils.toJsonNode(bom));
-        sbom.setType(SbomType.BUILD_TIME);
+        sbomerClient.updateSbom(String.valueOf(sbom.getId()), SbomUtils.toJsonNode(bom));
 
-        log.debug("SBOM: {}", sbom.toString());
+        log.info("SBOM '{}' updated with generated BOM!", sbom.getId());
 
-        log.info("Uploading generated SBOM...");
-
-        sbomerClient.save(sbom);
-
-        log.info("SBOM uploaded!");
+        return CommandLine.ExitCode.OK;
     }
+
+    protected abstract GeneratorImplementation getGeneratorType();
+
+    protected abstract Path generate();
 
     protected void doClone(String url, String tag, Path path, boolean force) {
         log.info("Cloning '{}' repository and '{}' tag...", url, tag);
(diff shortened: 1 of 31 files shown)


core/src/main/java/org/jboss/sbomer/core/enums/SbomStatus.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.sbomer.core.enums;
+
+/**
+ * Enum representing the status of the Sbom.
+ */
+public enum SbomStatus {
+    NEW, GENERATING, GENERATION_FAILED, PROCESSING, PROCESSING_FAILED, READY;
```
goldmann, line 24: This is pretty much up for a discussion, especially in case we would like to add multiple processing phases.

service/src/main/java/org/jboss/sbomer/model/Sbom.java
```diff
@@ -116,16 +117,22 @@ public class Sbom extends PanacheEntityBase {
     private String buildId;
 
     @Column(name = "root_purl")
-    @NotBlank(message = "Root purl identifier missing")
+    // @NotBlank(message = "Root purl identifier missing") // TODO
```
goldmann, line 120: So, I had to make it optional. Currently I change the approach and the SBOMs are saved by the service while not being fully populated with the actual Bom data. This is to support statuses of Sbom  resources, for examle: we don't have the sbom, because it's still generating and the status is `GENERATING`. And because of this, the purl will be empty at that stage.

core/src/main/java/org/jboss/sbomer/core/enums/SbomStatus.java
goldmann: The `*_FAILED` are not used yet. The reason for it is that we are not querying the TaskRun for status, nor we are receiving events from these which could led to updating the status. This will be changed at some point: https://github.com/project-ncl/sbomer/issues/41

goldmann:
@vibe13 I think this is now ready for review!

cli/src/main/java/org/jboss/sbomer/cli/commands/processor/AbstractBaseProcessCommand.java
```diff
@@ -65,31 +70,18 @@ public Integer call() throws Exception {
...
-    }
-
     protected abstract ProcessorImplementation getImplementationType();
 
     protected Bom doProcess(Bom bom) {
         log.info(
-                "Applying {} processing to the SBOM: {}",
+                "Applying {} processing to the SBOM: '{}'",
                 getImplementationType(),
                 bom.getMetadata().getComponent().getPurl());
```
vibe13, line 86: This log line might throw NPE, see LN 89, we need to handle the case where metadata or component are null

service/src/main/java/org/jboss/sbomer/service/SbomService.java
```diff
@@ -96,20 +116,40 @@ public void processSbom(Sbom sbom, ProcessorImplementation processor) {
...
      * @param processor Selected {@link ProcessorImplementation}
      */
-    public void processSbom(Long sbomId, ProcessorImplementation processor) {
-        processors.select(ProcessorLiteral.of(processor)).get().process(sbomId);
+    @Transactional
+    public Sbom process(Sbom sbom, ProcessorImplementation processor) {
+        log.debug("Preparing to process SBOM id '{}' with '{}' processor", sbom.getId(), processor);
+
+        // Create the child object
+        Sbom child = sbom.giveBirth();
```
vibe13, line 124: LOL

vibe13 (changes requested):

core/src/main/java/org/jboss/sbomer/core/enums/SbomStatus.java
goldmann: In my local work I changed this to: `NEW, IN_PROGRESS, FAILED, READY;` This fits much better with the work I do for updating the status based on the TaskRun outcome.

cli/src/main/java/org/jboss/sbomer/cli/commands/processor/AbstractBaseProcessCommand.java
goldmann: Good catch!

service/src/main/java/org/jboss/sbomer/service/SbomRepository.java
```diff
@@ -17,13 +17,17 @@
  */
 package org.jboss.sbomer.service;
 
+import java.util.Optional;
```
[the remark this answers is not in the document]
goldmann: Fixed!

service/src/main/java/org/jboss/sbomer/service/SbomRepository.java
```diff
@@ -17,13 +17,17 @@
...
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.transaction.Transactional;
 
 import org.jboss.sbomer.core.enums.GeneratorImplementation;
 import org.jboss.sbomer.core.enums.ProcessorImplementation;
 import org.jboss.sbomer.model.Sbom;
 
+import com.fasterxml.jackson.databind.JsonNode;
```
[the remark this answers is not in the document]
goldmann: Fixed!

cli/src/main/java/org/jboss/sbomer/cli/commands/generator/AbstractMavenBaseGenerateCommand.java
```diff
@@ -71,11 +70,14 @@ public abstract class AbstractMavenBaseGenerateCommand implements Callable<Integ
...
+        // First, fetch the SBOM metadata
+        Sbom sbom = sbomerClient.getById(parent.getParent().getSbomId());
+
         // Fetch build information
-        Build build = pncService.getBuild(parent.getParent().getBuildId());
+        Build build = pncService.getBuild(sbom.getBuildId());
 
         if (build == null) {
-            log.error("Could not fetch the PNC build with id '{}'", parent.getParent().getBuildId());
+            log.error("Could not fetch the PNC build with id '{}'", parent.getParent().getSbomId());
```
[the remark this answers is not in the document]
goldmann: Fixed!

cli/src/main/java/org/jboss/sbomer/cli/commands/processor/AbstractBaseProcessCommand.java
goldmann: Fixed!

core/src/main/java/org/jboss/sbomer/core/enums/SbomStatus.java
vibe13: Ok so `NEW` is as soon as a generation is requested, while `IN_PROGRESS`, `FAILED`, `READY` can refer both to generation and processing tasks, so e.g. to get a successful generated SBOM I would need to query for SBOMs with status = `READY` and generator != null and processor == null, while to get a successful processed SBOM I would need to query for SBOMs with status = `READY` and generator != null and processor != null ?

goldmann: Exactly what you said!

goldmann: With one small subtle change: `NEW` is both, for processing and generation. It's when the resource is created, but *not yet* worked on. In other words: after creation but before the TaskRun starts.

vibe13: ok cool, it makes it easier now. in case we will need multiple processing steps we might need to adjust, but many places in the code would need to change, so this would be a minor change. thanks!

vibe13 (approved):

PR #39, opened 2023-04-13
[NCL-7696] - Add rootPurl identifier to Sbom model and add endpoints

vibe13:
- Add rootPurl identifier to Sbom model
- Retrieve base and enriched sboms with buildId and rootPurl
- Add tests and fix test resources 
- Add a flag to enable or disable UMB consumers and producers

### DIFF
--- a/service/src/main/java/org/jboss/sbomer/rest/v1alpha1/SBOMResource.java
+++ b/service/src/main/java/org/jboss/sbomer/rest/v1alpha1/SBOMResource.java
@@ -32,6 +32,7 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
+import org.cyclonedx.model.Bom;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.media.Content;
 import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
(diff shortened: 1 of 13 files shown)


vibe13:
Hi @goldmann thanks for your suggestions, I like it much better and yes actually I should have thought of exposing both the sbom and the bom via REST, they will all be handy. I have pushed the suggested changes. Thanks!

service/src/main/java/org/jboss/sbomer/rest/v1alpha1/SBOMResource.java
```diff
@@ -140,11 +142,11 @@ public Sbom getById(@PathParam("id") String id) {
     }
 
     @GET
-    @Path("/build/{id}")
+    @Path("/build/{buildId}")
     @Operation(
             summary = "Get all SBOMs related to a PNC build",
             description = "Get all the SBOMs related to the specified PNC build, paginated.")
```
goldmann, line 148: I actually wonder if we should paginate it. Thoughts? Pagination is great if we are talking about UI, but for REST endpoints it may be unexpected. Thoughts?

vibe13: Maybe for this one we could avoid the pagination as there should not be a lot of SBOM associated with a build. But I would keep it for the general listAll endpoint.

goldmann (approved):

Issue #41, opened 2023-04-17
Propagate failed TaskRun statuses to resource statuses


goldmann:
Currently when a generation or processing fails, we are not updating the resource status.

PR #44, opened 2023-04-18
Implement automatic updates of statuses

goldmann:
This ensures that the status of the `Sbom` resource is always up to date according to the `TaskRun` status.

Fixes https://github.com/project-ncl/sbomer/issues/41


service/src/main/java/org/jboss/sbomer/k8s/TaskRunStatusHandler.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    protected void handleTaskRunUpdate(TaskRun taskRun) {
+        if (!isUpdateable(taskRun)) {
+            return;
+        }
+
+        SbomStatus status = toStatus(taskRun);
+
+        updateStatus(taskRun.getMetadata().getLabels().get(Constants.TEKTON_LABEL_SBOM_ID), status);
```
vibe13, line 208: Do we want to execute this if status is null? (could be returned by `toStatus`)

vibe13 (approved):
Just a minor comment but brilliant work Marek!!!!

service/src/main/java/org/jboss/sbomer/k8s/TaskRunStatusHandler.java
goldmann: Done!

vibe13 (approved):

PR #52, opened 2023-04-26
feat: Red Hat product processor and related refactoring

goldmann:
* allow for multiple proccessors to be executed for a single CLI run
* NCL-7791: created a temporary mapping that would help to populate the SBOM with the product-related information until we have it in PNC
* NCL-7794: implemented the product processor (`redhat-product`)
* refactoring of the processor codebase so that above changes are possible
* introduced mixin so that we can share same options
* some class renames
* introduced a different way to run CLI tests which should be more used in the future tests


cli/src/main/java/org/jboss/sbomer/cli/commands/processor/RedHatProductProcessCommand.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        BuildConfiguration buildConfig = pncService.getBuildConfig(build.getBuildConfigRevision().getId());
+
+        if (buildConfig == null) {
+            throw new ApplicationException(
+                    "BuildConfig related to the SBOM could not be found in PNC, interrupting processing");
+        }
+
+        ProductVersionMapping mapping = productVersionMapper.getMapping().get(buildConfig.getProductVersion().getId());
```
vibe13, line 97: ProductVersion might be null in the BuildConfigs

vibe13 (changes requested):
Great! Just a minor comment about a possible NPE

cli/src/main/java/org/jboss/sbomer/cli/commands/processor/RedHatProductProcessCommand.java
goldmann: Thanks, will update it!

goldmann: This should be now covered!

vibe13 (approved):

PR #56, opened 2023-04-28
Use managed PostgreSQL database on staging environment

vibe13:


service/src/main/resources/application.yaml
```diff
@@ -26,7 +26,7 @@ quarkus:
     info-title: SBOMer service
   hibernate-orm:
     database:
-      generation: update # TODO: This is for now, so that we can work on stage, remove this once we maintain the DB properly
+      generation: update
```
goldmann, line 29: Should it still be `update`?

goldmann (approved):

PR #58, opened 2023-04-30
Processing refactoring

goldmann:
First of all: this makes it possible to generate and process a Quarkus SBOM.

OK, this is a big one, since it involved changes in many places so that
things can work. Below is a summary of all the changes in this PR:

* Processing of a base generated SBOM is now controlled with the
	sbomer.processing.auto-process config parameter
* The configuration in sbomer.processing.default-processors controls
	what processors should be run by default for the automatic process
	as well as the sbom/[id]/process REST endpoint
* Better exception mappers, introduction of ClientException
* Unfortunately (hardcoding) of the bom.json, because for Quarkus Domino
	is prefixing the output file...
* Handling of artifacts that are Red Hat but not found in PNC (?!)
* Better logging configuration
* Update for Quarkus mapping from product version 377 to 382 (platform)
* Extended testing with WireMock, added some integration tests for CLI
* Update to the Sbom entity. Since we can now run multiple processors at
	one go it didn't make sense to have a single "processor" field. So,
	introduced a mapping so that many processors can be set. This has
	some implications over the SBOMer usage, because now we will not be
	able to fetch a single "enriched" Sbom, but I don't think this is
	a problem at all. We should work on Sbom id's eiher case. And because of this:
* Commenting out many endpoints that are not used now or are wrong after
	the change. Code was not removed / cleaned up yet, especially the queries.
* Extending of the configuration mapping into objects.
* Splitting SbomService and adding ProcessingService and GenerationService
* And finally: updating tests so that everything works.


goldmann:
For local testing you can use the `http://localhost:8080/api/v1alpha1/sboms/generate/build/AYL6AOH4DRQAA?generator=DOMINO`. This should generate the SBOM (and process it automatically) for `2.13.7.SP2-redhat-00001`.

service/src/main/resources/db-update/00001-init-1.0.0.sql
vibe13, line 1: This file need to be regenerated with a new patch script, will do in a following PR

service/src/main/java/org/jboss/sbomer/service/GenerationService.java
```diff
@@ -0,0 +1,131 @@
...
+        Sbom sbom = new Sbom();
+        sbom.setType(SbomType.BUILD_TIME); // TODO Is it always the case?
+        sbom.setBuildId(buildId);
+        sbom.setGenerator(generator);
+
+        // Store it in database
+        sbom = sbomService.save(sbom);
+
+        // Schedule the generation
+        generators.select(GeneratorLiteral.of(generator)).get().generate(sbom.getId());
```
vibe13, line 96: This is missing the generatorVersion and generatorArgs parameters, I will fix in a following PR

service/src/main/java/org/jboss/sbomer/model/Sbom.java
vibe13, line 1: All the named queries need to be fixed, they will cause errors when hibernate initializes the model

service/src/main/java/org/jboss/sbomer/rest/v1alpha1/SBOMResource.java
vibe13, line 1: I am ok in commenting them for now as we need to fix all the queries that now are not working anymore. Regardless of whether we are announcing those endpoints the in the outgoing notifications, we need them to retrieve information not based on SBOM ids, which might not be available, and not user friendly. Yes, we can create a query API to aggregate some of them, we can also bump the API version from `v1alpha1`

vibe13 (commented):
Ok so I think there are a few things that will need to be fixed before we have a working deployment on stage. Since you are off for a couple of days, I might merge this one and open a PR with the fixes to be made. Leaving open for now until I am also back from PTO tomorrow. Thanks for all this work!

service/src/main/java/org/jboss/sbomer/model/Sbom.java
goldmann: Ah, ok. Haven't realized it, didn't see it at least on development env.

service/src/main/java/org/jboss/sbomer/rest/v1alpha1/SBOMResource.java
goldmann: Yes, yes, we need some kind of other endpoints, it's just a little bit unclear how these would work with multiple processors. Related to this: I think we should allow regeneration of base SBOM and maybe use **latest processed** as the one that you would receive for a particular purl, for example. It would become handy in case of misconfigurations in SBOM generation requests.

The query API would be nice, I think. It would give the client some flexiblity.

service/src/main/java/org/jboss/sbomer/service/GenerationService.java
goldmann: Ha, good catch!

vibe13 (approved):
Ok I will approve and merge, and then apply some fixes

PR #64, opened 2023-05-05
Enable the automatic notifications of completed SBOM

vibe13:


service/src/main/java/org/jboss/sbomer/features/umb/producer/NotificationService.java
```diff
@@ -50,7 +47,7 @@
 /**
  * Service implementation responsible for the notification of completed SBOMs.
  *
- * @author Marek Goldmann
+ * @author Andrea Vibelli
```
goldmann, line 50: 😎

goldmann (approved):

PR #65, opened 2023-05-05
Fix validation of UMB message, due to schema file not being found at …

vibe13:
…runtime


goldmann (approved):

goldmann:
Thanks a lot! I suspected this can happen :)

PR #70, opened 2023-05-11
[NCL-7818] Unable to generate again when previous generation failed

vibe13:
- Created new feature config section to handle cleanup of successful taskruns and retry of failed taskruns
- Modify the taskruns handler to cleanup completed successful runs (if configured), store the final message of completed runs into the Sbom, and retry failed runs (if enabled)
- Update tests to verify cleanup and retries


service/src/main/java/org/jboss/sbomer/model/Sbom.java
```diff
@@ -165,6 +166,12 @@ public class Sbom extends PanacheEntityBase {
     @ToString.Exclude
     private JsonNode sbom;
 
+    @JsonIgnore
+    @Lob
+    @Type(type = "org.hibernate.type.TextType")
+    @Column(name = "task_run_msg")
+    private String taskRunMsg;
```
goldmann, line 173: I guess nitpicking, but I would not make this name related to TaskRun at all. For example calling it `message` or `statusMessage` would be better IMHO.

vibe13: right, looks much better

goldmann (approved):

PR #79, opened 2023-05-24
[NCL-7867] Improve logging with MDC

vibe13:
1. Lower log level of RSQL operations
2. Add MDCUtil class and add required header to the generated REST client (the header name is "fixed" because it is automatically handled by existing utilities)
3. Add required dependencies for JSON logging and MDC utilities
4. Handle the addition and removal of MDC in the code
5. Add Tekton Metadata label to keep the buildId information with the ability to add it to the MDC

@goldmann : 

- Because of 2, I know that the signature of the PNC Client is not that clean anymore, but it is the only way I found to add headers with the generated REST client
- Because of 3, log is not beautiful anymore but the json formatting is required on all our services to be ingested properly by our internal logging systems
- Because of 5, I had to alter the method signature of the AbstractTektonTaskRunner class and implementations. If you prefer to keep the same method signature, I can keep it unchanged but fetch the `buildId` information with a call to the DB.

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/client/SBOMerClient.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/client/SBOMerClient.java
@@ -22,6 +22,7 @@ import javax.validation.Valid;
 import javax.ws.rs.BeanParam;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
+import javax.ws.rs.HeaderParam;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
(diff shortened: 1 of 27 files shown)


goldmann (approved):
I'm not that worried about the change in the signature at this moment. I think we should do whatever is needed, because I'm going to propose some changes to the codebase either case that will affect how things are structured and executed.

Thanks a lot for the changes, you rock!

ACK

vibe13:
Thanks a lot @goldmann , yes the code is evolving rapidly, I expect changes to come constantly :-). Thanks!

PR #120, opened 2023-06-30
Set generation request failed reason according to generation outcome

goldmann:
https://issues.redhat.com/browse/NCL-7994

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateCommand.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateCommand.java
@@ -28,6 +28,7 @@ import org.jboss.sbomer.cli.feature.sbom.command.PathConverter;
 import org.jboss.sbomer.core.errors.ApplicationException;
 import org.jboss.sbomer.core.features.sbom.config.runtime.Config;
 import org.jboss.sbomer.core.features.sbom.config.runtime.ProductConfig;
+import org.jboss.sbomer.core.features.sbom.enums.GenerationResult;
 import org.jboss.sbomer.core.features.sbom.utils.MDCUtils;
 import org.jboss.sbomer.core.features.sbom.utils.ObjectMapperProvider;
 
@@ -88,14 +89,14 @@ public class GenerateCommand implements Callable<Integer> {
 
         if (configPath == null) {
             log.info("Configuration path is null, cannot do any generation.");
-            return 2;
+            return GenerationResult.ERR_CONFIG_MISSING.getCode();
         }
 
         log.info("Reading configuration file from '{}'", configPath.toAbsolutePath());
 
         if (!Files.exists(configPath)) {
             log.info("Configuration file '{}' does not exist", configPath.toAbsolutePath());
-            return 2;
+            return GenerationResult.ERR_CONFIG_MISSING.getCode();
         }
 
         // It is able to read both: JSON and YAML config files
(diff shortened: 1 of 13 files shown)


goldmann:
image

goldmann:
```json
{
	"id": "B0814E9273FC4CE",
	"creationTime": "2023-07-04T07:56:09.232218Z",
	"status": "FAILED",
	"buildId": "AZPKTTSU2HQAA",
	"config": {
		"buildId": "AZPKTTSU2HQAA",
		"products": [
			{
				"generator": {
					"args": "--batch-mode",
					"type": "maven-cyclonedx",
					"version": "2.7.9"
				},
				"processors": [
					{
						"type": "default"
					},
					{
						"type": "redhat-product",
						"errata": {
							"productName": "RHTESTPRODUCT",
							"productVariant": "8Base-RHTESTPRODUCT-1.1",
							"productVersion": "RHEL-8-RHTESTPRODUCT-1.1"
						}
					}
				]
			}
		],
		"apiVersion": "sbomer.jboss.org/v1alpha1"
	},
	"reason": "Generation failed. Product with index '0' (TaskRun 'sbom-request-azpkttsu2hqaa-hqfrw-1-generate-0') failed: system failure. See logs for more information."
}
```

service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/SbomGenerationRequest.java
```diff
@@ -92,7 +92,7 @@ public class SbomGenerationRequest extends PanacheEntityBase {
     @ToString.Exclude
     private JsonNode config;
 
-    @Column(name = "reason", nullable = true, updatable = true)
+    @Column(name = "reason", nullable = true, updatable = true, length = 1024)
```
vibe13, line 95: I would add  `@Type(type = "org.hibernate.type.TextType")` and do not bother with max size? Unless we then make sure we trim the reason to be max 1024 chars when we create it. Btw we already set "text" type in stage and prod, so no worries about the length already.

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateCommand.java
```diff
@@ -90,7 +99,20 @@ public Integer call() throws Exception {
...
 
         // It is able to read both: JSON and YAML config files
-        Config config = objectMapper.readValue(configPath.toAbsolutePath().toFile(), Config.class);
+        Config config;
+
+        try {
+            config = objectMapper.readValue(configPath.toAbsolutePath().toFile(), Config.class);
+        } catch (StreamReadException e) {
+            log.error("Unable to parse the configuration file", e);
+            return 2;
```
vibe13, line 108: Wouldn't be bad to have the statuses mapped internally with constants maybe? Just for improved visibility.

vibe13 (changes requested):
Just left a couple of small comments, overall great improvement. Thanks!!!

service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/SbomGenerationRequest.java
goldmann: Good catch, will adjust!

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateCommand.java
goldmann: Yes, I think it makes sense!

I was also thinking about storing the exit code in the database as well so that we could query based on the status. If we would add on top of this some kind of enum, then we would have even a nice failure reason in the entity, a description in text format and this would be in a single place.

vibe13: Yes good idea! Anything which will make our life easier to query and debug is surely welcome, talking for experience with PNC which has so many use cases. Go for it! :-)

goldmann:
image

I introduced a new field within the `ConfigMap` and the `SbomGenerationRequest` entity: `result`. This includes the enum value that represents the reason for the failure in a way that could be queried.

In case of multiple failures, we set the failure to `ERR_MULTI`.

goldmann:
I forgot to update the DB scripts.

goldmann:
And fixed.

vibe13 (approved):
Great addition! Approved :-D

PR #121, opened 2023-07-03
Fix potential NPEs.

matejonnet:


service/src/main/java/org/jboss/sbomer/service/feature/sbom/k8s/reconciler/GenerationRequestReconciler.java
```diff
@@ -147,24 +147,27 @@ public GenerationRequestReconciler() {
...
+                                    }
+                                },
+                                () -> new IllegalStateException("Finished task should return non-empty value.")
+                            );
+                    } else {
+                        generationRequest.setStatus(SbomGenerationStatus.INITIALIZING);
+                    }
+                    return UpdateControl.updateResource(generationRequest);
+                })
+            .orElseGet(() -> UpdateControl.noUpdate());
```
goldmann, line 170: I have mixed feelings about this. It became much more hard to read IMHO. @vibe13 WDYT?

service/src/main/java/org/jboss/sbomer/service/feature/sbom/k8s/reconciler/GenerationRequestReconciler.java
```diff
@@ -420,44 +420,19 @@ public UpdateControl<GenerationRequest> reconcile(
...
-            return UpdateControl.noUpdate();
-        }
+        UpdateControl<GenerationRequest> action = switch (generationRequest.getStatus()) {
+            case NEW -> reconcileNew(generationRequest, secondaryResources);
+            case INITIALIZING -> reconcileInitializing(generationRequest, secondaryResources);
+            case INITIALIZED -> reconcileInitialized(generationRequest, secondaryResources);
+            case GENERATING -> reconcileGenerating(generationRequest, secondaryResources);
+            case FINISHED -> reconcileFinished(generationRequest, secondaryResources);
+            case FAILED -> reconcileFailed(generationRequest, secondaryResources);
+        };
```
goldmann, line 435: I like this one! 👍🏻

service/src/main/java/org/jboss/sbomer/service/feature/sbom/k8s/reconciler/GenerationRequestReconciler.java
vibe13: Mmmmm wow yes I have a very hard time reading the code as well.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/k8s/reconciler/GenerationRequestReconciler.java
vibe13: Nice one! We still need the 
```
        // This would be unexpected.
        if (action == null) {
            log.error(
                    "Unknown status received: '{}'' for GenerationRequest '{}",
                    generationRequest.getStatus(),
                    generationRequest.getMetadata().getName());
            return UpdateControl.noUpdate();
        }
```
part, correct?

vibe13 (changes requested):
I am not convinced about the refactoring of the first part because I will spend many minutes reading it, every time I will need. 
I like the new switch syntax, but I think we still need to check whether action == null.

goldmann:
Closing this one, it's hanging there for a bit too long. I think I'll probably incorporate some of the changes suggested here i na subsequent PR.

PR #126, opened 2023-07-05
feat: support for cleaning up resources when generation is finished

goldmann:
Cleanup is controlled in application (service) settings. When enabled the Kubernetes resource as well as all dependent resources are removed from the cluster. Additionally any leftovers on the volume are cleaned too.

Removal is done no matter if the status is FINISED or FAILED. Resources are synced with the DB earlier, plus we have logs available in a separate system so we can debug everything at any time.

https://issues.redhat.com/browse/NCL-7964


goldmann:
One more thing: resources are cleaned up by default in the production profile. In development these are not, so that we can debug failures easily.

vibe13 (approved):
Oh I like this one! No more leftovers left in our clusters! :-). Big +1!

PR #129, opened 2023-07-07
fix: fixes for the chart found at staging deployment

goldmann:


goldmann:
@vibe13 FYI - merged.

PR #130, opened 2023-07-10
docs: update docs, mostly for Helm

goldmann:


vibe13 (approved):

docs/deployment.md
```diff
@@ -9,27 +9,30 @@ Once this is done, the automatic deployment to staging begins. This deployment u
...
+```bash
 oc create sa sbomer-deployer
 ```
 
+NOTE: The _ServiceAccount_ name may differ between [environments](environments.md).
+
 ### Required RoleBindings
 
-By default no additional RoleBindings are required.
+A *ClusterRole* `admin` should be applied to this *ServiceAccount* for a given namespace.
```
matejonnet, line 28: Can you describe also how to apply a ClusterRole to a ServiceAccount.

goldmann: I'm a bit unsure how to deal with his section in general. The SA and RoleBindings are already created on the target (stage/prod) envs. These won't change. For development environment we don't need these.

Thoughts?

PR #132, opened 2023-07-10
[NCL-7987] - Reintroduce the optimizations for NO_REBUILD_REQUIRED builds

vibe13:
So, I have made some optimizations here.
- I have re-introduces the SBOMerClient in the CLI, so that is is able to query the service for existing SBOMs or SBOMGenerationRequests
- I have added a facade to easier consume the Response from the SBOMerClient
- I had to work on the RSQLParser so it can handle the rsql query on our Enums

- In the AbstractGenerateCommand, in case of NO_REBUILD_REQUIRED builds:

1. I search for the latest successful SBOMGenerationRequests for the same `buildId`;
2. In case there is a result, I compare the current ProductConfig with the ProductConfig stored for the SBOMGenerationRequest found above, having the same ProductIndex
3. If the current ProductConfig is the same of the ProductConfig found above (the generation was done with the same config parameters), I search the SBOMs generated by the SBOMGenerationRequest found above (having the same ProductIndex)
4. In case I found a matching SBOM, I do not regenerate it from scratch but I skip the generation and reuse it, removing the Errata properties in case they need to be updated
5. In all other cases (no successful SBOMGenerationRequests found for the same buildId, different ProductConfig used to generate the past SBOMs; no matching SBOMs found) --> I do the generation as usual
6. Additionally, to optimize not only the generation phase but also the processing phase, in case the generated SBOM has already the required enriched properties for a given component (`pnc-build-id` and `pnc-environment-image`), I do not query PNC. This is useful in the above scenario, where the reused SBOMs has already all the enrichment


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/client/facade/SBOMerClientFacade.java
```diff
@@ -0,0 +1,124 @@
...
+        };
+        try {
+            Page<SbomGenerationRequest> sbomRequests = ObjectMapperProvider.json().readValue(json, typeReference);
+            if (sbomRequests.getTotalHits() > 0) {
+                return new ArrayList<>(sbomRequests.getContent());
+            }
+        } catch (JsonProcessingException e) {
+            log.warn("Could not find existing successful SBOM Generation Requests for PNC build '{}'", buildId, e);
+        }
+        return null;
```
goldmann, line 78: Maybe better return empty list?

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/client/facade/SBOMerClientFacade.java
```diff
@@ -0,0 +1,124 @@
...
+        };
+        try {
+            Page<Sbom> sboms = ObjectMapperProvider.json().readValue(json, typeReference);
+            if (sboms.getTotalHits() > 0) {
+                return new ArrayList<>(sboms.getContent());
+            }
+        } catch (JsonProcessingException e) {
+            log.warn("Could not find SBOMs for Generation Request '{}'", requestId, e);
+        }
+        return null;
```
goldmann, line 112: Similarly to above, maybe empty list?

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/AbstractGenerateCommand.java
```diff
@@ -95,30 +141,128 @@ public Integer call() throws Exception {
         // Get the correct scm information for builds which have either SUCCESS or NO_REBUILD_REQUIRED status
         String scmUrl = build.getScmUrl();
         String scmTag = build.getScmTag();
+
+        JsonNode bom = null;
+
         if (org.jboss.pnc.enums.BuildStatus.NO_REBUILD_REQUIRED.equals(build.getStatus())) {
```
goldmann, line 147: Shouldn't we check here whether `force`  is set and if it is, the generate it either case, without even looking at the possible `NO_REBUILD_REQUIRED` state?

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/client/facade/SBOMerClientFacade.java
vibe13: +1

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/AbstractGenerateCommand.java
vibe13: Mmmmm I don't think the `force` and `state` are correlated. The `force` is at the moment used to wipe any existing workspace in case is already exists, before the clone operation. I honestly fail to think of how that could happen, given that a new pod is spin up to take care of that for every build? The NO_REBUILD_REQUIRED means than no new build was done at all in PNC, because there were no changes to the build config. So ideally, it would be kind of wrong to regenerate the SBOM from scratch. If the force (as intended in PNC) was used to force a new build anyway, we would not receive the NO_REBUILD_REQUIRED state.

goldmann: OK.

Note about the CLI:

I guess the think is that you need to start thinking about CLI being separate from the service. I should be able to run the CLI, without service. This was the idea. Service is a wrapper/orchestrator around the CLI, but I should be able to do everything in the service, just locally.

goldmann (approved):

PR #134, opened 2023-07-10
Populate SBOMER_SERVICE_URL env variable

vibe13:
The cli has this config:
```
    "sbomer":
      url: "${SBOMER_SERVICE_URL:http://localhost:8080}"
```
which is not populated anymore. This PR aims to repopulate the service url so it can be called from the CLI via the facade.


goldmann:
I don't think this will work. You are setting it on the service deployment, but it's used in the CLI, which is run in the context of the TaskRun. There we cannot just depend on the route, because in development deployment route doesn't exist (plus, we would need to honor the `service.route.enabled`). It looks like it's a candidate for a new template in `templates/_helpers.tpl` (see `sbomer.serviceAccountName` ).

vibe13:
@goldmann I made some changes, can you help me verify if that is ok? I have run `helm template --force  sbomer ./helm/` and the value is there, but running `helm template  --values ./helm/env/prod.yaml sbomer ./helm/` afterwards did not seem to have changes applied. Thanks!

helm/templates/service/deployment.yaml
```diff
@@ -76,6 +76,8 @@ spec:
               value: {{ .Values.env }}
             - name: SBOMER_ROUTE_HOST
               value: {{ .Values.service.route.host }}
+            - name: SBOMER_SERVICE_URL
+              value: "https://{{ .Values.service.route.host }}"
```
goldmann, line 80: Since we have it in the config map, then why not something like this?

```suggestion
              valueFrom:
                configMapKeyRef:
                  name: {{ .Release.Name }}-config
                  key: SBOMER_SERVICE_URL
```

vibe13: ah! I forgot to remove it .... not needed in the service deployment I believe

goldmann: But I guess it should be exposed in Tekton Tasks.

goldmann (approved):

PR #139, opened 2023-07-18
Remove 'from-maven' domino subcommand

aloubyansky:
This subcommand has been deprecated a long time ago. Domino figures out whether it's pointed at a Maven or Gradle project and proceeds with the dependency analysis accordingly.


aloubyansky:
There is no actual difference between `from-maven report` and `report`, it's the same `report` command.

vibe13 (approved):
Thanks Alexey for pointing this out!

PR #141, opened 2023-07-23
[NCL-7976] - Add Brew component information in case it is available

vibe13:


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/DefaultProcessCommand.java
```diff
@@ -126,6 +141,54 @@ private void processBuild(Component component, Build build) {
         SbomUtils.addPedigreeCommit(component, build.getScmUrl() + "#" + build.getScmTag(), build.getScmRevision());
     }
 
+    private void processBrewBuild(Component component, Artifact artifact) {
+        log.debug("Looking up artifact '{}' in Brew...", artifact.getPublicUrl());
```
goldmann, line 145: I would like to have a log statement when the artifact in Brew:

a) was found
b) wasn't found

This will help debug "close" this log message with an outcome.

cli/src/main/resources/application.yaml
```diff
@@ -47,6 +47,9 @@ sbomer:
     api-url: "${SBOMER_CACHE_SERVICE_URL:${PNC_SERVICE_HOST:orch.psi.redhat.com}}"
     product-mapping:
       environment: "${SBOMER_PRODUCT_MAPPING_ENV:prod}"
+  koji:
+    hub-url: "https://brewhub.engineering.redhat.com/brewhub"
+    web-url: "https://brewweb.engineering.redhat.com/brew"
```
goldmann, line 52: The configuration refactoring unified the way we define server addresses. I based on using the `sbomer.[NAME].host` pattern. In this case it would be: `sbomer.koji.host`.

So, I would set it to followng:

```yaml
sbomer:
  koji:
    hub-url: "https://${sbomer.koji.host}/brewhub"
    web-url: "https://${sbomer.koji.host}/brew"
```

Afterwards, in the `task-generate.yaml` file I would add following env variable:

```yaml
- name: SBOMER_KOJI_HOST
  value: {{ .Values.koji.host }}
```

And finally, add the hostname to the sbomer-support repo we have, for all environments.

```yaml
koji:
  host: "THEHOSTNAME"
```

(Actually, I've done the last part for ya already, a MR is open)

P.S This of course implies that you rebase 😀

goldmann (changes requested):

goldmann (approved):

PR #144, opened 2023-07-24
feat(rest): sort by default on creationTime in desc order (newest first)

goldmann:


goldmann:
I'll probably need to fix tests :)

goldmann:
Aaaaand done.

I found it useful to get newest entities first. @vibe13 do you agree?

vibe13 (approved):
Yes sounds good!

PR #150, opened 2023-08-03
feat: introduce a new generator framework

goldmann:
This should help write tests and make execution of them faster.

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -93,6 +93,23 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-inline</artifactId>
+      <version>4.11.0</version>
+      <scope>test</scope>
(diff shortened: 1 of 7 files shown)


cli/pom.xml
```diff
@@ -92,6 +92,23 @@
...
+    <dependency>
+      <groupId>io.github.hakky54</groupId>
+      <artifactId>logcaptor</artifactId>
+      <version>2.9.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <version>2.0.7</version>
```
goldmann, line 110: We are using some old slf4j version currently. It was required to upgrade it so that we can test log statements properly. I imagine this manual update won't be required once we move to Quarkus 3.

vibe13 (approved):
Good work!

PR #192, opened 2023-08-08
chore(dep): removal of custom jansi version

goldmann:


goldmann:
Related: #179.

@vibe13 Do you remember why we added custom Jansi version? I tested the CLI with this removal and it works fine.

goldmann:
Ok, probably because of the `Caused by: java.lang.ClassNotFoundException: org.fusesource.jansi.AnsiOutputStream` error.

goldmann:
Since we upgraded to Quarkus 3, we can now merge this.

vibe13:
@goldmann I have missed the old message but yes, that is exactly why I had to use a custom jansi version, a class was relocated and I had issues with the build finder.

vibe13 (approved):
Ok so let's remove the custom version now! :-) Thanks!

PR #207, opened 2023-09-06
chore: version unification

goldmann:
We refrain from defining the Quarkus version explicitly now. Instead we use the transient version as defined by the Operator SDK Quarkus extension. This means that we are now at Quarkus 3.2.4.Final.

Unification of naming of properties related to versions was done as well.

Removal of custom Hibernate types. It's not needed anymore in Quarkus 3 which uses Hibernate 6.2.


vibe13 (approved):
Ah good finding! Looks good and cleaner now, thanks!

PR #214, opened 2023-09-07
Fix certificates and update logging

goldmann:


vibe13 (approved):
go go go!! :-) Thanks!

PR #216, opened 2023-09-09
feat: add support for retrieving product version from group config

goldmann:
In case the build config used to build the project doesn't have the `productVersion` set, we will try to get this information from a group configuration this build config is assigned to (if any).

In case the build config is assigned to multiple group configs -- we will take into account all of them. This will result in multiple "products" in SBOMer terms which will rpoduce multiple SBOMs for all of the productVersions found.

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/service/PncService.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/service/PncService.java
@@ -17,7 +17,10 @@
  */
 package org.jboss.sbomer.cli.feature.sbom.service;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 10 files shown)


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/service/PncService.java
```diff
@@ -128,50 +138,110 @@ public BuildConfiguration getBuildConfig(String buildConfigId) {
...
         }
 
         log.debug("BuildConfig: {}", buildConfig);
 
         ProductVersionRef productVersion = buildConfig.getProductVersion();
 
-        if (productVersion == null) {
+        // Product version was found, let's use it!
+        if (productVersion != null) {
+            return List.of(productVersion);
```
vibe13, line 209: I am now wondering: should it be exclusive? Meaning, if a buildConfig has a productVersion assigned, should we get anyway all the productVersions linked to any groupConfig, the buildConfig is part of?

goldmann: Like I mentioned, I don't necessarily like the idea of having to support multiple product versions because of the assignments to multiple build configs. The line you highlight is basically preserving old behavior; if we have it explicitly set, then let's use it without asking questions. I was looking at the group config approach as a fallback, but again -- I'm not sure I like it at all. I would probably better make a requirement for SBOMer to work to assign a product version to a BC.

vibe13: Ok , so we can keep the old behavior (use the Product Version assigned to a BC) and fall back to the GroupConfigs, so the PR looks good. I don't actually dislike this approach because there are different ways the teams are using BC and GC, so we are covering more use cases. Thanks!

vibe13 (approved):

PR #221, opened 2023-09-14
[NCL-8180] Allow to provide configuration when requesting generation

goldmann:
This change makes it possible to provide the Config with the product and
generator configuration as part of the REST API call to
/api/v1alpha1/sboms/generate/build/{buildId}. It enables providing a
custom configuration for that particular generation.

In case the configuration is provided it will be used, which means that
the initialziation phase which includes crafrting configuration won't be
run at all.

This feature is very useful for cases where we want to try out different
configurations to "get it right", for example when onboarding someone.
Another use case is to adjust configuration so that SBOM can be
generated for an old build.

Configuration can be sent in the body as YAML (application/yaml
content type) or JSON (application/json content type). Provided
configuration is validated and adjusted if necessary (putting default
values, if not provided).


goldmann:
Will mark this as a draft, will extend it with tests.

core/src/main/java/org/jboss/sbomer/core/config/ConfigSchemaValidator.java
```diff
@@ -0,0 +1,74 @@
...
+        try {
+            InputStream is = SchemaValidator.class.getClassLoader().getResourceAsStream("schemas/config.json");
+            schema = new String(is.readAllBytes(), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new ApplicationException("Could not read the configuration file schema", e);
+        }
+
+        try {
+            String writeValueAsString = ObjectMapperProvider.json().writeValueAsString(config);
+            System.out.println(writeValueAsString);
```
vibe13, line 62: Leftover or needed?

vibe13 (approved):
@goldmann great work, which will be handy to test out very easily new builds! Thanks!

core/src/main/java/org/jboss/sbomer/core/config/ConfigSchemaValidator.java
goldmann: Leftover!

goldmann: Removed!

vibe13 (approved):

aloubyansky:
Is this feature documented somewhere?
In case of passing custom config, would the resulting manifest be persisted?
Is there an option to generate a config w/o storing it permanently?

goldmann:
1. I'll give you the link shortly.
2. Yes.
3. No.

PR #235, opened 2023-09-27
Add oidc auth to cli clients communicating with pnc

jomrazek:


jomrazek:
once we're ready also https://github.com/project-ncl/sbomer/blob/main/service/src/main/resources/application-prod.yaml#L52 will need to be reconfigured to service account

thescouser89 (approved):
LGTM

goldmann:
> LGTM. Does sbomer perform write / delete actions on PNC that would require the token though?

We only read content from PNC API.

goldmann:
@vibe13 Could you please look at this PR?

jomrazek:
I understood it probably like future-proofing, but it is there so we converted it. Also I believe Andrea is on meetings this week.

cli/src/main/resources/application.yaml
```diff
@@ -41,6 +41,12 @@ quarkus:
     console:
       format: "%d{HH:mm:ss,SSS} %-5p [%c{3.}] (%t) %s%e mdc:[%X]%n"
       stderr: true
+  oidc-client:
+    enabled: true
+    auth-server-url: http://localhost:8180/auth/realms/quarkus
```
vibe13, line 46: @jomrazek This needs to be configured to be our internal SSO right? We have private files for this, but double checking it needs to be configured. Thanks!

jomrazek: Yes on internal config files this will be configured to RedHat internal SSO system

vibe13 (approved):

PR #237, opened 2023-09-27
Always use https for PNC-Orch url

thescouser89:


goldmann:
I guess this change would require to setup WireMock to be served via https. Other option, probably better, would be to use `http` schema for test profile and `https` for everything else.

```java
ConfigProvider.getConfig().unwrap(SmallRyeConfig.class).getProfiles()
```

Could be handy.

PR #238, opened 2023-09-28
[NCL-8018] - Add SBOM generation for Gradle builds 

vibe13:
@goldmann 
With the following changes, I am adding the capability to generate CycloneDX SBOMs for Gradle builds. 

The new generation uses an init-script, to be used from the CycloneDX Gradle plugin when doing the generation, to configure properly the repositories to be used when fetching dependencies (similar to Maven settings.xml) and to configure the CycloneDX parameters used for the SBOM generation.

As Gradle depends heavily on the Gradle and JDK installed in the `sbomer-generator` container  (where we default to Java17, which is not compatible with earlier versions of Gradle thus making the SBOM generation impossible), I also had to make some changes to the TaskRuns.
I have added an intermediate `sbomer-env-config` TaskRun, to be run after the `sbomer-init` TaskRun, which fetches the build environment information used to generate a PNC build. The environment properties are stored inside the DB, and passed over to the `sbomer-generate` TaskRun which is responsible to trigger the SBOM generation. These environment information are then used to call a script which detects (and returns) the best matching software version installed in the container (the managed software types are `java`, `gradle`, `maven` for now). The best matching software version is then configured to be used prior the call to the SBOM Generation. For example, when having to generate an SBOM for a build which was built using `Gradle 5.6.2 and JDK 11`, these steps allow to search for any best matching Gradle / JDK installed locally, set them as default, and then call the generation, which otherwise would have failed.

Why did I create a new intermediate `sbomer-env-config` TaskRun and not used the `sbomer-init`? Because I did not feel like the users would need to care about the environment information. This way, if they provide a Product Config (thus letting us skip the `sbomer-init` TaskRun), they can still benefit transparently from the environment setup.

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateEnvConfigCommand.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateEnvConfigCommand.java
@@ -76,20 +76,14 @@ public class GenerateEnvConfigCommand implements Callable<Integer> {
     /**
      * Retrieves environment configuration from a PNC Build, compliant with SDKMan.
      *
-     * @return a Map object if the configuration could be retrieved or {@code null} otherwise.
+     * @return a Map object if the configuration could be retrieved or an empty map otherwise.
      */
-    private Map<String, String> environmentConfig() {
+    private Map<String, String> environmentConfig(Build build) {
         // Make sure there is no context
         MDCUtils.removeContext();
-        MDCUtils.addBuildContext(buildId);
+        MDCUtils.addBuildContext(build.getId());
         log.debug("Attempting to fetch environment configuration from a PNC build");
 
-        Build build = pncService.getBuild(buildId);
-
-        if (build == null) {
-            log.warn("Could not retrieve PNC build '{}'", buildId);
-            return Collections.emptyMap();
-        }
         Map<String, String> buildEnvAttributes = build.getEnvironment().getAttributes();
         if (buildEnvAttributes == null || buildEnvAttributes.isEmpty()) {
             log.debug("Build environment attributes not found for the specified PNC build");
@@ -106,12 +100,20 @@ public class GenerateEnvConfigCommand implements Callable<Integer> {
 
     /**
      *
-     * @return {@code 0} in case the environment config file was generated successfully. An empty Map is valid as
-     *         default versions will be used
+     * @return {@code 0} in case the environment config file was generated successfully or {@code 1} in case the
+     *         provided buildId could not be found. A returned empty Map is valid as default versions will be used
      */
     @Override
     public Integer call() throws Exception {
-        Map<String, String> envConfig = environmentConfig();
+
+        Build build = pncService.getBuild(buildId);
+
+        if (build == null) {
+            log.warn("Could not retrieve PNC build '{}'", buildId);
+            return GenerationResult.ERR_GENERAL.getCode();
+        }
+
+        Map<String, String> envConfig = environmentConfig(build);
 
         if (envConfig.isEmpty()) {
             log.debug(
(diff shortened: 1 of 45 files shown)


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateEnvConfigCommand.java
```diff
@@ -0,0 +1,147 @@
...
+        // Make sure there is no context
+        MDCUtils.removeContext();
+        MDCUtils.addBuildContext(buildId);
+        log.debug("Attempting to fetch environment configuration from a PNC build");
+
+        Build build = pncService.getBuild(buildId);
+
+        if (build == null) {
+            log.warn("Could not retrieve PNC build '{}'", buildId);
+            return Collections.emptyMap();
```
goldmann, line 91: See comment below, I would check for the build earlier and probably pass it to this method.

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateEnvConfigCommand.java
```diff
@@ -0,0 +1,147 @@
...
+    @Inject
+    PncService pncService;
+
+    @Inject
+    ConfigReader configReader;
+
+    /**
+     * Retrieves environment configuration from a PNC Build, compliant with SDKMan.
+     *
+     * @return a Map object if the configuration could be retrieved or {@code null} otherwise.
```
goldmann, line 79: Probably not `null` by an empty Map :)

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateEnvConfigCommand.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     *
+     * @return {@code 0} in case the environment config file was generated successfully. An empty Map is valid as
+     *         default versions will be used
+     */
+    @Override
+    public Integer call() throws Exception {
+        Map<String, String> envConfig = environmentConfig();
```
goldmann, line 114: Maybe we could add a check whether the build exists and fail the execution. Non-existing build and the inability to fetch environment configuration are treated the same by the code.

images/sbomer-generator/cyclonedx-init.gradle
```diff
@@ -0,0 +1,80 @@
...
+
+initscript {
+  repositories {
+    maven {
+        url "https://plugins.gradle.org/m2/"
+    }
+  }
+
+  dependencies {
+    classpath "org.cyclonedx:cyclonedx-gradle-plugin:1.7.4"
```
goldmann, line 54: Are we able to control the version via the SBOMer config?

goldmann (changes requested):
This is a great feature! Thanks for working on it, I see it required a lot of work. I don't have many comments, just small things, everything else looks sane!

I'm not supper happy about the other TaskRun execution, but I get the reasoning behind it.

There are some test failures that need to be fixed, as you mentioned.

Either case, great job!

images/sbomer-generator/cyclonedx-init.gradle
vibe13: I initially spent some time on this and failed, but will look again . Yes I wanted to make it configurable

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/auto/GenerateEnvConfigCommand.java
vibe13: good point, changed the above!

images/sbomer-generator/cyclonedx-init.gradle
goldmann: I'm fine for it being hardcoded in the first pass! I just saw the config mentioning the version and was curious how it works.

vibe13:
@goldmann Hey Marek! I have added some more tests, and fixed the one failing. We could either merge this now if you are happy with it, or keep it open for some more days until I make the tooling version of the Gradle plugin configurable as per the above comment. If we merge it now, I will open a new PR next week for that. Thanks !

goldmann (approved):

goldmann:
I approved it, like I said -- happy to work incrementally on adding support for this very welcome feature.

Thanks a lot for this contribution!

vibe13:
Thank you for the review! Merging then :-D. Will surely make some more improvements to this!

PR #243, opened 2023-10-10
feat(service): NCL-8202: skip sending manifests from SBOM paged results

goldmann:
The `/sboms` endpoint which allows for searching for SBOM entities in a paginated way a DTO was introduced to limit the amount of data that is read and sent to the client. In most cases, the user is not interested in the SBOM content at this point, because it's more about checking whether a given result exists. Once the client found the result a request to `sboms/ID` can be initiated (or `/sboms/ID/bom` to get the SBOM content only) to read the full object.

Change summary:

* Introduction of DTOs represented as Java Records
* Introduction of the `v1alpha2` API, which becomes now the default, meaning that it will be announced in the UMB messages sent after finished generation.
* The dev and test modes include now sample entries in the database, so that when the application boots in these modes, the data will be already available for querying.
* Refactoring of the RSQL handling due to the requirements to build the criterias for DTO's (`CriteriaAwareRepository`). There are still more things to be done to make it much prettier.
* SQL statements are logged in tests as well as in dev mode.


service/src/main/java/org/jboss/sbomer/service/feature/sbom/rest/v1alpha1/SBOMResource.java
```diff
@@ -29,6 +29,7 @@
 import org.jboss.sbomer.core.SchemaValidator.ValidationResult;
 import org.jboss.sbomer.core.config.ConfigSchemaValidator;
 import org.jboss.sbomer.core.config.SbomerConfigProvider;
+import org.jboss.sbomer.core.dto.v1alpha2.SbomRecord;
```
vibe13, line 32: not required here

service/src/main/resources/application-prod.yaml
```diff
@@ -17,7 +17,7 @@
 #
 
 sbomer:
-  api-url: "https://${SBOMER_ROUTE_HOST}/api/v1alpha1/"
+  api-url: "https://${SBOMER_ROUTE_HOST}/api/1alpha2/"
```
vibe13, line 20: typo: `v1alpha2`

service/src/main/java/org/jboss/sbomer/service/feature/sbom/rest/criteria/AbstractCriteriaAwareRepository.java
```diff
@@ -0,0 +1,141 @@
...
+
+    protected <X> CriteriaQuery<X> handleSort(CriteriaQuery<X> query, Root<T> root, String sort) {
+        if (Strings.isEmpty(sort)) {
+            return query;
+        }
+
+        String compliantSort = CustomPredicateSortBuilder.rsqlParserCompliantSort(sort);
+        log.debug("Modified RSQL sort string from: '{}' to a RSQL parser compliant format: '{}'", sort, compliantSort);
+
+        CustomizedJpaPredicateSortVisitor<?> sortVisitor = new CustomizedJpaPredicateSortVisitor<>(Sbom.class)
```
vibe13, line 132: This is probably not an `Sbom.class` but a more generic `entityType` ? So it can be used by all entity classes?

vibe13 (changes requested):
This is a great feature Marek! And I like how you simplified a few classes which were a bit too complex. This v1alpha2 endpoint will really make our searches easier!
I left a couple of comments for some fixes. Thanks!

service/src/main/java/org/jboss/sbomer/service/feature/sbom/rest/criteria/AbstractCriteriaAwareRepository.java
goldmann: Good catch, I think I was using it for testing. It should be here: `entityType` instead.

goldmann:
@vibe13 Thanks! There are still many areas that could be improved. I have a feeling that all the criteria builder could be revamped. I was also thinking about making it a separate library, that could be used across our projects (NCL, SBOMer, younameit) so that we don't repeat ourselves. Plus, it would make the `javax` -> `jakarta` transition smoother, maybe?

PR #246, opened 2023-10-10
Make the gradle cyclonedx plugin version configurable with config options

vibe13:
@goldmann Ok, here I make the plugin version configurable via config options. I might in future add some logic to override the version based on the Gradle version required, as it looks like older Gradle versions (4.10) does not work with versions 1.7.x of the plugin. But with this PR, it will be possible to specify the version via configuration.

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/GradleCycloneDxGenerateCommand.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/GradleCycloneDxGenerateCommand.java
@@ -43,10 +43,7 @@ public class GradleCycloneDxGenerateCommand extends AbstractGradleGenerateComman
     @Override
     protected Path doGenerate(String buildCmdOptions) {
         ProcessBuilder processBuilder = new ProcessBuilder().inheritIO();
-
-        // Making the plugin version configurable
-        processBuilder.command()
-                .add(String.format("PLUGIN_VERSION=%s", toolVersion()));
+        processBuilder.environment().put("PLUGIN_VERSION", toolVersion());
 
         List.of(buildCmdOptions.split(SPLIT_BY_SPACE_HONORING_SINGLE_AND_DOUBLE_QUOTES))
                 .stream()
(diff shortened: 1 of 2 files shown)


vibe13:
retest this please

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/GradleCycloneDxGenerateCommand.java
```diff
@@ -44,6 +44,10 @@ public class GradleCycloneDxGenerateCommand extends AbstractGradleGenerateComman
     protected Path doGenerate(String buildCmdOptions) {
         ProcessBuilder processBuilder = new ProcessBuilder().inheritIO();
 
+        // Making the plugin version configurable
+        processBuilder.command()
+                .add(String.format("PLUGIN_VERSION=%s", toolVersion()));
```
goldmann, line 49: Didn't you want `environment()` here?

vibe13: this is the Gradle CycloneDX plugin version, which is specified in:
```      
gradle-cyclonedx:
        default-version: "1.7.4"
```
or using the `--tool-version` parameter.

goldmann: I think my comment was more about using the `command()` method. This adds the passed parameter to the **command** that is being executed, but from what I see you want to set an environment variable, for which I would expect using `environment().put("PLUGIN_VERSION", toolVersion())`.

vibe13: ok right, I was going for a simpler way, to execute the command like `PLUGIN_VERSION=1.7.4 gradle etc etc `

goldmann: Aaaah. I wasn't sure if this was possible with `ProcessBuilder`. If it is, I'm fine with it!

goldmann (approved):

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/GradleCycloneDxGenerateCommand.java
vibe13: right, you are correct, let's not abuse the command and use the proper way!

goldmann (approved):

PR #259, opened 2023-10-20
fix(service): handle more UMB errors and force the reconnection

goldmann:
This adds handling for errors that are thrown by the system. In case of a failure, it is logged (as error, with the exception as well) and the flag of being connected is set to false. I was thinking whether this is the best approach or not.

I don't think the proposed logic is perfect. It forces reconnection on any JMS exception, but I haven't figured out (yet) more accurate representation of fatal errors.

### DIFF
--- a/e2e/src/test/java/org/jboss/sbomer/test/e2e/rw/StageGenerationRequestIT.java
+++ b/e2e/src/test/java/org/jboss/sbomer/test/e2e/rw/StageGenerationRequestIT.java
@@ -148,7 +148,7 @@ public class StageGenerationRequestIT extends E2EStageBase {
         });
     }
 
-    //@Test
+    // @Test
     @Order(6)
     public void ensureUmbMessageWasSentForGradle4Build() {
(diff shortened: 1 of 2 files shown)


goldmann:
After reading the [javadocs for `setExceptionListener`](https://docs.oracle.com/javaee%2F7%2Fapi%2F%2F/javax/jms/JMSContext.html#setExceptionListener-javax.jms.ExceptionListener-) it looks like it is catching serious errors.

There are some things to consider:

1. Should we forcibly reconnect in as part of `onException` logic or just set the `connected` boolean to false (and letting the cron job to reconnect)?
2. If we agree to reconnect, should we be cleaning the connection at the moment of catching the error?

vibe13:
Hi  @goldmann thanks for the PR! I think we can rely on the periodic check and reconnection, the queue will retained and messages should not be lost anyway. If it turns up to be still an issue, we can revisit the strategy. Thanks!

PR #261, opened 2023-10-23
fix(oidc): ensure dev mode works after introducing oidc client

goldmann:
This makes the OIDC client disabled for dev mode.

Additionally upgrade Tekton to be in sync with OpenShift.


vibe13 (approved):
Nice fix! Thanks Marek!

PR #265, opened 2023-10-24
feat(service): NCL-8201: exposure of /sboms/purl/[PURL] endpoint

goldmann:
This new endpoint makes it possible to retrieve the SBOM by giving just the purl. In case of multiple SBOMs will be found for a given purl, latest SBOM will be returned.

In case no SBOM can be found, 404 will be returned.

Additionally `/sboms/purl/[PURL]/bom` is exposed as well to retrieve the BOM content.


goldmann:
Please note that this was added to the `v1alpha2` API version. I consider `v1alpha1` frozen from now on.

vibe13 (approved):
Ok those endpoints look familiar :-D from the first implementations. Looks good to me! Thanks Marek!

goldmann:
Yah, true. These are basically a convenience methods. As explained on the ticket you can do the same thing with the current API, but it would require two calls instead of one.

PR #270, opened 2023-10-30
[NCL-8263] Enhance purl for all Red Hat components with repository_url

goldmann:
This change updates the processing so that in case of a Red Hat component the purl will be enhanced with the `repository_url` pointing to https://maven.repository.redhat.com/ga.

Community components wont have this adjustment applied.

Additionally, the other commit updates the properties so that OIDC really works in dev mode :)


goldmann:
This won't be needed, but I'll open another PR with just the OIDC changes.

PR #280, opened 2023-11-24
Upgrade to domino 0.0.100

aloubyansky:
* record dependency graphs instead of trees (in line with the CycloneDX Maven plugin)
* use CycloneDX spec 1.5

`--include-non-managed` is enabled by default when `--manifest` is present.


goldmann (changes requested):
Could you please update this line as well?

https://github.com/project-ncl/sbomer/blob/65a0e7fd1028f652981812fefc1615fea39e23f2/images/sbomer-generator/install.sh#L62

goldmann (approved):

PR #285, opened 2023-12-04
fix(api): page generation requests properly

goldmann:
When returning generation requests pages, do the paging correctly. Currently it fails because we try to  count Sbom entities instead of SbomGenerationRequests.


vibe13 (approved):
Nice finding!!!

PR #286, opened 2023-12-04
feat(NCL-8317): add periodic reconnects

goldmann:
Add periodic reconnects in case no new messages were received over particular period of time. Currently it is configured for allowing up to 3 hours of quiet period. If more than this is discovered, a client reconnection is scheduled.


vibe13 (approved):
LGTM! Finger crossed :-D

PR #293, opened 2023-12-08
NCL-8326: use reactive handling of messages by default

goldmann:
This will remove the old Qpid handling of messages as well.


goldmann:
Merged it myself, going to test it on stage.

PR #324, opened 2024-01-14
NCL-8352: Documentation improvements

goldmann:
This switches documentation from Markdown files into Asciidoc which makes it possible to write much richer documentation. Along with this change we are going to use Antora (https://antora.org/) to generate and manage our documentation.

Some new scripts were added, mosti importantly for now: `./hack/docs-preview.sh`.

Generation (and preview) requires some software to be installed. Assuming you have node installed you should run: `npm ci` which should install everything needed.


vibe13 (approved):
Massive! Thanks Marek, looks great!

PR #335, opened 2024-01-25
Introduction of GitHub actions and GitLab Pages workflow

goldmann:
This alters how we perform CI and CD. Basically this is a move out of Jenkins and make use of two other services: GitHub Actions and GitLab Pipelines.

On GH we do the build process, this includes the service and cli code as well as all images. This is run for every PR as well as every merge to `main` branch. The difference is that when merged, these images are pushed to quay. This is where GH Actions stop.

This merge event will be picked up by GitLab at some point (delay up to 5 minutes). This will start the deployment process on staging environment. Once this is done, the e2e tests will be run and if successful, a there will be a proposal to move to production. Deploying to production requires a manual approval, at least for now.


goldmann:
Note: this still requires implementing running e2e tests, I'll do this shortly.

goldmann:
Support for running e2e tests was added.

PR #353, opened 2024-01-31
Improvements for cache to make it HA

goldmann:
This ensures that we can run the cache with multiple replicas. Each cache has it's own storage and session affinity is configured so that we don't hit random replicas for a given client. Cache invalidation was updated so that content in the cache is stored for one _month_ instead of one _minute_, how it was before...


vibe13 (approved):
Nice improvement! Thanks!

PR #360, opened 2024-02-08
feat(SBOMER-16): do not bundle tools

goldmann:
This is an attempt to make the generator image smaller and more agile. Instead of bundling all the necessary tools and SDKs at build time, we do it selectively at runtime based on the environment information as well as config.

The only thing we install at build time is: JDK 17 which is required to run SBOMer CLI. This will prpobably change in the future once we move to natives.

We install nvm as well sdkman to be able to install required tools at build time.

Fixes: https://issues.redhat.com/browse/SBOMER-16


goldmann:
Generally, this PR cleans the `task-generate.yaml` file. It moves the content into separate bash files divided by use cases for better maintenance.

The parameters sent to the task are not JSON formatted. It was YAML before.

We make now use of `jq` to parse it and get relevant data based on which we install appropriate tools.

Overall, this PR should decrease the size of the image from about 4 GB to about 900 MB.

images/sbomer-generator/runtime/run.sh
```diff
@@ -0,0 +1,108 @@
...
+    echo "About to install '${tool}' with version '${version}'"
+
+    case "$tool" in
+    "java")
+        # Special case, we have only major version
+        install_major_java "${version}"
+        echo "java=${version}" >>"${HOME}/.sdkmanrc"
+        ;;
+    "node")
+        install_nodejs "${version}"
```
vibe13, line 65: In Node.js case we installed v18.19.0 and v20.11.0, using the latest one (v20.11.0) for all the generations, to handle incompatibility issues with the plugin. See comment in `sdk_nvm_find_closest_installed_software.sh LN#140`

vibe13 (changes requested):

vibe13 (dismissed):

goldmann:
I've rebased it on top of current main and going to merge it now.

PR #362, opened 2024-02-12
SBOMER-14: ingest all builds

goldmann:
This commit makes it possible to generate manifests for any PNC builds, no matter if these are related to a product, or not. For products that do not have a provided config, we generate a default one based on the environment which was used at the build time. The generator type and generator parameters are the default ones for the particular environment combination.

Red Hat product processor is optional from now on.

Additionally this moves the logic for environment configuration preparation into the initialization phase. Removes the env-detection phase entirely.

The database is cleaned up and the `env_config` column is now dropped, data is migrated into `config`.

https://issues.redhat.com/browse/SBOMER-14

https://issues.redhat.com/browse/SBOMER-19
https://issues.redhat.com/browse/SBOMER-15
https://issues.redhat.com/browse/SBOMER-20


vibe13 (approved):
Massive work @goldmann ! +10000

PR #363, opened 2024-02-13
fix(SBOMER-12): otel fixes

goldmann:
Updates configuration to use new format.

Helm values are exposed via Quarkus-readable envs now.

https://issues.redhat.com/browse/SBOMER-12


goldmann:
This change requires update to the vault for the endpoint so it can work properly!

goldmann:
This was updated, please see ticket for more info.

vibe13 (approved):

PR #380, opened 2024-02-19
Tool installation fixes

goldmann:
This fixes a problem where SDKMAN! wasn't run at all (the tool selection didn't include anything to install), but we always wanted to initialize it.

Second is an enhancement and bugfix in one. When there are issues installing tools we will retry it up to 3 times. Better error handling was implemented as well.


vibe13 (approved):
Great job with retries!

PR #381, opened 2024-02-22
feat(SBOMER-49): improvements in error handling

goldmann:
There are two changes that are both related to error handling:

1. The SBOMer CLI returning appropriate exit codes that make it possible to understand what went wrong by just looking at the exit code.
    * Adds new exception classes that are mapped to exit codes
    * Adds more tests for CLI
    * Introduction of `SBOMER_CLI_FAILURE_REASON_PATH` which (if set) will
      make SBOMer CLI populate it with a JSON content providing more
      information on why the execution failed. Initial implementation, may
      change. Example:
      
        ```json
        {
          "message": "Could not retrieve PNC build 'aaaA4WLAOY3BJIAA'",
          "exitCode": 33
        }
        ```
2. SBOMer service API improvements like unified error messages and proper JSON output always.
    * API: A unified `ErrorResponse` object is return in any type of error. we provide more information in cases we can.
    * API: Introduction of `AbstractExceptionMapper` class to simplify handling of errors. Probably can be still improved.
    * API representation was updated to that Swagger presents schemas and examples properly.


vibe13 (approved):
This is nice indeed and will provide much more clarity about the errors! And good that you abstracted the SbomResource, something was looking at in my previous PR but did not want to add too much to it. Thanks Marek!

goldmann:
This PR is not ready yet for a re-review. I need to go through it one more time tomorrow.

goldmann:
Ok, this should be ready now for re-review.

I was trying to make the API definition reuse the endpoints that we already have, but I failed many times to do it clean, whether to Swagger limitations (I wanted to have it nicely presented in Swagger) or by OpenApi annotation shading. Nothing really worked well. Then I started to think that actually, the duplication is expected, because we are defining entirely new API version. So, I decided to reuse the code but make the definition separate for each API version. This will make it easy to understand the API (from a developer POV), hopefully.

Generally the idea is now that in case we introduce new API version, we create new version package (`service/src/main/java/org/jboss/sbomer/service/feature/sbom/rest`) and an ApiProvider there, new mapper (`service/src/main/java/org/jboss/sbomer/service/feature/sbom/mapper`,) new DTOs (`core/src/main/java/org/jboss/sbomer/core/dto`). When we start a new version we basically copy the version - 1 and make adjustments. We should **refrain from reusing DTO's between versions**. This will become messy easily!

WDYT?

vibe13:
@goldmann The service is still evolving rapidly but I don't expect many API versions changes once it is settled. I am fine now with some duplication, let's see how bad it becomes.

vibe13 (approved):

goldmann:
The #415 took over it.

PR #383, opened 2024-02-26
feat(SBOMER-51): use the Jakarta-supported version of PNC REST library

goldmann:
Earlier we had to add many of the PNC REST library, because this was the easiest way to just have it to be able to upgrade to Quarkus 3 which switched from javax to jakarta. Now, since we have a dedicated artifact from the PNC team to support Jakarta we can get rid of these.


goldmann:
Marking it as a draft until PNC 2.7 will be out.

goldmann:
And thanks to @michalovjan we now have a version we can consume!

goldmann:
Closes #417

PR #384, opened 2024-02-26
[SBOMER-37] Add SBOM generation for Deliverable Analyzer analysis operations

vibe13:
This PR adds the manifesting capabilities to SBOMer for the PNC Operations of type "Deliverable Analysis'.
The Deliverable Analysis consists of a list of files identified inside the analyzed zips / packages (belonging to a Product Milestone); if the files were found (built from source) in internal build systems, additional metadata is returned.

This feature leverages all the data returned from a Deliverable Analysis operation, and creates a SBOM manifest for them, enriching the content of the base SBOM with internal metadata. A new custom Generator (`maven-cyclonedx-operation`) has been implemented for this workflow, and will produce CycloneDX compliant manifests. This generator takes also care of the processing of the SBOM because there is no need to run `default` nor `redhat-product` processors since all the information needed for the enrichments is already available from the Deliverable Analysis.

The model has been adapted to handle this different generations: the specific `buildId` inside the `Sbom` and `SbomGenerationRequest` entities now have a generic `identifier` (which now contains the buildId and operationId identifiers), and the `SbomGenerationRequest` has an additional `type`, which can be `GenerationRequestType.BUILD` or `GenerationRequestType.OPERATION`.

The `GenerationRequestReconciler` has been extended to also handle the new OperationInit and OperationGeneration tasks, but with the new JOSDK version it's possible to handle the same resource with different reconciler; this will be refactored in a future PR.

The UMB has also been extended to handle the incoming messages related to DeliverableAnalysis operation completions, which now trigger the creation of the manifest; new UMB notification will also be sent to notify the creation of such new manifests.

This work is related to https://issues.redhat.com/browse/SBOMER-37.

@goldmann Happy to discuss any improvements or refactoring about this work, there will surely be but I wanted this to land as it has taken long already.


service/src/main/java/org/jboss/sbomer/service/feature/sbom/rest/v1alpha1/SBOMResource.java
goldmann, line 1: Changing the `Sbom` and `SbomGenerationRequest` entities will change the API output for this version. I don't think we want it.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/rest/v1alpha2/SBOMResource.java
goldmann, line 1: Similar to the `v1alpha1` version -- we will violate the contract of not changing the API in case we modify `Sbom` and `SbomGenerationRequest` entities.

goldmann (changes requested):
Oh man, this was a lot of work!

I've made comments inline which are focused on specific bits. Below are general comments. I'm not sure how to tackle this, but let's start chatting about it and we will see where we will end up.

- I'm not really fond of the _operation_ naming. If this would be used internally only, then I think it would be OK, but we expose this to users with the CLI and REST API. I don't have any proposal at this stage.
- We need to bump the API version to `v1alpha3`. We change the structure of the content we send to the client, for example in `SbomGenerationRequest` `buildId` becomes `identifier` and the `config` element being something different depending on the type. Similar thing applies to `Sbom` entity.
- Why do we have `MAVEN_CYCLONEDX_OPERATION`? From what I see the _operation_ is not tied to Maven at all. Am I missing something?
- UMB operation finished message: I would probably adjust the current message: the `build` would be set to `null` in case of operation. Introduction of new `trigger` (or something else) element that would encapsulate a `build` and an `operation` with a property to distinguish between these two. Or something similar.
- I'm thinking about the requirement to have a second workflow for the _operation_ type. What I mean is that now we have second set of Tekton Tasks, CLI commands and so on. I'm wondering why we don't have this logic embedded within the already existing classes. Sure, there are differences, but the process seems to be very similar. This is related to the comment above about UMB message differences and in general, the `config` content in the `SbomGenerationRequest` class. I have a feeling that this could be covered using already existing constructs. Yes, this would mean that we need to generalize things and this is exactly what I was planning. We are too tied to PNC now. The introduction of `identifier` in the `SbomGenerationRequest` is a right step! But maybe it should be even part of the config itself? Which then would be depending on the type of work that needs to be done (PNC Maven build, PNC Gradle Build, PNC _operation_, Custom ZIP, and so on).

**Not really related to this PR**, but I'm thinking whether we should the initialization phase entirely as part of the controller work and not spinning up a Tekton Task. More like: metadata generation – do it in controller, real work: in Tekton Task. We would have much more control over it. Downside: more load on the controller, but not sure if we should worry about it much.

vibe13:
Thanks @goldmann  for the review and feedback!



> I'm not really fond of the _operation_ naming. If this would be used internally only, then I think it would be OK, but we expose this to users with the CLI and REST API. I don't have any proposal at this stage.

So you don't know how much I thought about it also :joy:. Don't really like `operation` either. At the end we are going to generate manifests for files, so should we use `package` instead, or something similar?

> We need to bump the API version to `v1alpha3`. We change the structure of the content we send to the client, for example in `SbomGenerationRequest` `buildId` becomes `identifier` and the `config` element being something different depending on the type. Similar thing applies to `Sbom` entity.

Yes good point.

> Why do we have `MAVEN_CYCLONEDX_OPERATION`? From what I see the _operation_ is not tied to Maven at all. Am I missing something?

No, it's just because I initially started with the idea of creating intermediate POM files representing the ZIPs, and feed them to the standard Maven CycloneDX plugin. I realized I was just creating more and more issues to myself, so I implemented my own manifest creation using the data I already had. So we can definitely remove the `MAVEN` suffix from that name (`CYCLONEDX_PACKAGE`)?

> UMB operation finished message: I would probably adjust the current message: the `build` would be set to `null` in case of operation. Introduction of new `trigger` (or something else) element that would encapsulate a `build` and an `operation` with a property to distinguish between these two. Or something similar.

Ok. We might need to double check if the new format won't break anything on the services listening for them...considering also that they will have low resourcing. But sure, the 2 messages smelled a lot of duplication.

> I'm thinking about the requirement to have a second workflow for the _operation_ type. What I mean is that now we have second set of Tekton Tasks, CLI commands and so on. I'm wondering why we don't have this logic embedded within the already existing classes. Sure, there are differences, but the process seems to be very similar. This is related to the comment above about UMB message differences and in general, the `config` content in the `SbomGenerationRequest` class. I have a feeling that this could be covered using already existing constructs. Yes, this would mean that we need to generalize things and this is exactly what I was planning. We are too tied to PNC now. The introduction of `identifier` in the `SbomGenerationRequest` is a right step! But maybe it should be even part of the config itself? Which then would be depending on the type of work that needs to be done (PNC Maven build, PNC Gradle Build, PNC _operation_, Custom ZIP, and so on).

We can refactor for example the `generate` command to accept an `identifier` instead of a `buildId` or an `operationId`, and additionally a `type` to run different generations. Similarly for the Tekton tasks, we can probably assume there will always be an  `init` phase followed by a `generation` phase. I did not want to push this additional refactoring here. WRT the identifier in the config itself... actually this worries me a bit. It would mean no indexing on the DB (and we want to have a LOT of SBOMs:-) ), needing a deserialization of the config (unless we index the config directly and use json native DB functions) and a bit more pain when querying in the DB.

I will address the specific comments on the code tomorrow. Thanks!

> **Not really related to this PR**, but I'm thinking whether we should the initialization phase entirely as part of the controller work and not spinning up a Tekton Task. More like: metadata generation – do it in controller, real work: in Tekton Task. We would have much more control over it. Downside: more load on the controller, but not sure if we should worry about it much.

The heavy lifting is surely done in the `generation` phase, so we could think about the above. In the `init` phase we need to call a few endpoints but the load should not be something worrying I guess.

goldmann:
Alright, it looks like your PR triggered inside me some thinking. Originally it was a reply to your comment, but then I started diverging and come up with what you can read below.

Let's think about the `GeneratorType`, https://github.com/project-ncl/sbomer/blob/1e1952c3c011173791bd40624f205929c91f3c43/core/src/main/java/org/jboss/sbomer/core/features/sbom/enums/GeneratorType.java. Currently what it describes is the type of the output format (CycloneDX) and the build tool (Maven, Gradle).

We are probably missing a few things.

The only supported source of content is PNC currently. This made some impact on how we designed the interaction with the tool/service and structures internally. For example we require to provide a PNC `build-id` (now we try to generalize it). We heavily depend on PNC and, as long at it makes sense currently, I think it makes things more problematic long-term. I would like to decouple this long-term.

We have following ingredients:

- Output format (currently only CycloneDX, but who knows what it could be in the future)
- Type of content which we want to process (source code, delAn, random zip, container)
  - In case we process source code: build tool used and source code location

Our current implementation is pretty much focused on:

- Source code being the source for information
- PNC being the provider of the content

How to make it more generic?

As a user I have some knowledge about what I want to do. The first this is the _format_ of the content I want to generate manifest for. This determines other things:

- `SOURCE` -- we use the source code to generate the manifest (most common use case):
	- Providers (how to get the source code):
		- `GIT`: provide repository coordinates to check out
		- `LOCAL`: a path to a directory containing source code
		- `PNC`: a build identifier that will help us obtain the source code
	  - Generators – basically what is the generator that supports a given build tool:
		  - `GRADLE_PLUGIN`
		  - `MAVEN_PLUGIN`
		  - `MAVEN_DOMINO`
- (not implemented) `DIST` – a distribution which would be some packaged content (zip, tar, etc)
	- Providers:
		- `URL` – a link to a distribution
- (not implemented) `IMAGE` – a container image
	- Providers:
		- `REGISTRY` – coordinates to an image in a registry
- (this PR basically ) `DEL_AN` – Deliverables Analyzer service output: very specific use case; no providers (we fetch everything from PNC) and there is a single generator (again, all the content comes from PNC).

Current implementation is purely a `SOURCE` format of content with the `PNC` provider and generators listed above. And that's it. It does the job, but it's not very flexible. This PR changes this a bit.

Output of the generation is a manifest. This manifest may need to to be post-processed to add more information to it from systems that we have. But only some of these generators may require post-processing. The generator would influence what processors can be applied to the generated manifests. So, instead of a hard coded list we would need have it defined per generator.

In our current wording, the `GRADLE_*` and `MAVEN_*` generators for `SOURCE` format would use the `DEFAULT` processor, but `DEL_AN` would not, because we already have the information.

Last bit is the understanding whether this is a manifest of a product build. This could be another processor. We would probably need to have more specific processors compared to what we have now. `REDHAT_PRODUCT` should become something like `ERRATA` or `ERRATA_PRODUCT` processor. Or a generic `PROPERTY` processor that would add arbitrary content to the main component.

And the last thing is the format. We can have `CYCLONEDX` as the the internal and default (and currently only) output format. We probably need some room to change things like this in the future as well.

vibe13:
Ok I like the idea of making this more generic. I would like to have a generic `generate` endpoint which accepts a json config of different formats and then triggers the generation. No need to have the `build` / `operation` details visible in the endpoint. We can probably start creating  a few different config options based on the `format`, validate them, and trigger the specific generators and processors.

> Output of the generation is a manifest. This manifest may need to to be post-processed to add more information to it from systems that we have. But only some of these generators may require post-processing. The generator would influence what processors can be applied to the generated manifests. So, instead of a hard coded list we would need have it defined per generator.

Mmmm I am not sure I follow this. The generators are more about _how_ the manifests are generated, while the processors are more related to _what_ has been generated and needs to be post-processed. So I tend to associate the processors to the `format` and `provider` (e.g. `SOURCE` + `PNC` would need to run the `DEFAULT` processor regardless of the generator used (so no matter if it was Maven / Gradle / NPM etc).

 
> In our current wording, the `GRADLE_*` and `MAVEN_*` generators for `SOURCE` format would use the `DEFAULT` processor, but `DEL_AN` would not, because we already have the information.

See above. I think that is not related much with the Gradle / Maven bit, but more about that is was a `SOURCE` + `PNC`?

> Last bit is the understanding whether this is a manifest of a product build. This could be another processor. We would probably need to have more specific processors compared to what we have now. `REDHAT_PRODUCT` should become something like `ERRATA` or `ERRATA_PRODUCT` processor. Or a generic `PROPERTY` processor that would add arbitrary content to the main component.

Although I would love to have the `PROPERTY` one with a list of arbitrary properties to add,  we need to validate that for example the Errata properties would have to be 3 and with specific names. Need to think a bit more the best option.

> 
> And the last thing is the format. We can have `CYCLONEDX` as the the internal and default (and currently only) output format. We probably need some room to change things like this in the future as well.

Definitely. We can start specifying the format version as a start for CycloneDX. We are defaulting to spec v1.4 value, but we need to make it configurable.

So now the million dollar question: what about this PR? Do we want / address any of the ^ in this PR?

goldmann:
I can answer the $$$ question: **no, we don't want to change it this significantly**. I would suggest that we do necessary changes (for me the blocker is the API change, this needs to be addressed). I'll go over it again and see if there is anything else. I'll try to keep it concise this time :)

Oh, and I'll probably reply to your comments above as well. Tomorrow :)

goldmann:
> Mmmm I am not sure I follow this. The generators are more about how the manifests are generated, while the processors are more related to what has been generated and needs to be post-processed. So I tend to associate the processors to the format and provider (e.g. SOURCE + PNC would need to run the DEFAULT processor regardless of the generator used (so no matter if it was Maven / Gradle / NPM etc).

I think I introduced a few concepts without really explaining these: format, provider and generator. _Generator_ is something that takes the _format_ delivered by a _provider_ and transforms it into a manifest.

_Format_ is basically a bucket that describes the nature of the _thing_, for example `SOURCE`, or `DIST`. _Provider_ is implementation on how to obtain the thing in a given format.  And _generator_ is the real workhorse which takes the provided content and generates a manifest.

Regarding your comment - I think we are basically both right. And I would probably even make a statement: the format determines the processors: generally besides `DEL_AN` everything requires some type of adjustments. But the "source" is not generating the manifest. Generators are. This means that generators need to be aware of processors that need to be run. Hence we are both right :)

Some other points while I was thinking about it:

- The _default_ processor (or the idea, at least) is something that we apply always. I don't think this will change over time -- we want always to improve the quality of the manifests by adding as much information as we can. We don't let clients disable this processor. Which brigs me to next point:
- I have a feeling that processors shouldn't be really configurable for end users. We don't allow it now (we always add the _default_ processor), but it's still part of the runtime configuration that can be provided by the client. This means that we kind of mix responsibilities of the runtime configuration between SBOMer and client, which is bad. There should be probably a different config for the user from the one we use internally (runtime config).

> Although I would love to have the PROPERTY one with a list of arbitrary properties to add, we need to validate that for example the Errata properties would have to be 3 and with specific names. Need to think a bit more the best option.

I wonder if we need to validate it. Shouldn't it be the client responsibility? We could explain the process and current requirements and that's it. How this is used would depend on the client.

Nice side feature: adding more data to manifest's properties would be easy!

vibe13:
@goldmann Hey Marek I have addressed a few issues and mainly kept the api backward compatibility. Let me know what else I should address here and what we can deal with later. Thanks!

goldmann (approved):

goldmann:
OMG. That's huge. Approved!

vibe13:
Thanks! Merging then! :-)

PR #388, opened 2024-03-04
feat(SBOMER-53): initialize configuration always

goldmann:
iRun the initialization always so that we don't end up with semi-initialized configuration files which coudl happen in cases where config files are provided by end users.

https://issues.redhat.com/browse/SBOMER-53


vibe13 (approved):
Nice, LGTM

vibe13 (approved):

PR #415, opened 2024-04-03
feat(SBOMER-49): improvements in error handling

goldmann:
There are two changes that are both related to error handling:

1. The SBOMer CLI returning appropriate exit codes that make it possible to understand what went wrong by just looking at the exit code.
    * Adds new exception classes that are mapped to exit codes
    * Adds more tests for CLI
    * Introduction of `SBOMER_CLI_FAILURE_REASON_PATH` which (if set) will
      make SBOMer CLI populate it with a JSON content providing more
      information on why the execution failed. Initial implementation, may
      change. Example:
      
        ```json
        {
          "message": "Could not retrieve PNC build 'aaaA4WLAOY3BJIAA'",
          "exitCode": 33
        }
        ```
2. SBOMer service API improvements like unified error messages and proper JSON output always.
    * API: A unified `ErrorResponse` object is return in any type of error. we provide more information in cases we can.
    * API: Introduction of `AbstractExceptionMapper` class to simplify handling of errors. Probably can be still improved.
    * API representation was updated to that Swagger presents schemas and examples properly.

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/CLI.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/CLI.java
@@ -23,6 +23,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.Callable;
 
+import org.jboss.sbomer.cli.errors.SbomerExitCodeExceptionMapper;
+
 import jakarta.enterprise.context.control.ActivateRequestContext;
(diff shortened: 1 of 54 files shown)


goldmann:
Supersedes #381.

vibe13 (approved):

PR #450, opened 2024-04-24
S3 log storage and logging improvements

goldmann:
This PR implements S3 log storage. This is NOT replacing the standard log capturing for stdout, it's an additional feature.

Subprocess execution was reworked as well so that we use a single `ProcessRunner` class to maintain every execution. This simplifies it and makes it more stable. This change wraps the output of the  subprocesses which makes it easier to parse because these will be JSON-formatted.

https://issues.redhat.com/browse/SBOMER-8


vibe13 (approved):
Nice feature Marek! Big +1 :-)

PR #473, opened 2024-05-09
[SBOMER-29] - Expose API endpoint to request SBOM generation for output from DelAnalyzer

vibe13:
This replaces https://github.com/project-ncl/sbomer/pull/471.

@goldmann This PR is meant to expose a new REST API to trigger a new Deliverable Analysis operation in PNC.

- The REST API will create an entry in the DB containing the Product Config metadata and the identifier of the new operation created in PNC via the PncService. It will not create any ConfigMap because there would be nothing to do anyway, as the creation of the new analysis in PNC will send back a UMB notification once finished. Until then, there is nothing to do
- Once a UMB notification is received, if there is a pending requests in the DB which a matching operation Id, it will be used to create a new ConfigMap with the same id, otherwise a new ConfigMap will be created.


goldmann (approved):

aloubyansky:
Thanks @vibe13! Is there a little doc describing the endpoint and how to use it?

PR #494, opened 2024-06-05
fix(SBOMER-86): handle deliverable analyzer failures

goldmann:
This makes it possible to handle messages from PNC for operations that have failed. Previously we kind of assumed these will be successfull, but we were proved wrong very fast.

When we receive a failure message, we update the Generation request entity as well and set the proper status with a reason pointing what went wrong.

https://issues.redhat.com/browse/SBOMER-86


vibe13:
Hi @goldmann thanks for the PR. There is one issue, that the msg body does not contain the result :-(
```
 "msg": {
    "attribute": "deliverable-analysis-state-change",
    "deliverablesUrls": [
      "https://XXX/my-zip.zip"
    ],
    "milestoneId": "2852",
    "operationId": "BAQZQCCUXYIAA",
    "status": "FINISHED"
  },
```
But you could query PNC via rest for the operation (`pnc-rest/v2/operations/deliverable-analyzer/BAQZQCCUXYIAA`) and get the result from there, until we can update the msg sent from PNC.

goldmann:
Oh,  riiiight,  I was looking at the REST result. I'll update this!

goldmann:
I updated the PR with the additional check.

vibe13 (approved):
Good stuff!

PR #572, opened 2024-07-04
refactor: generalize Generation configuration

goldmann:
This change makes implementing different types of configurations much easier. Configs are different for different generation types. In the early days we had just one related to PNC build, but now we have 4, which will probably rise in the future as well.

In order to make implementation of newa features easier configuration handling had to be reworked. A Config class with helper methods was introduced and now all configs are derived from it. This makes serialization and deserialization much easier. Besides this we can now operatore on the object level and not JsonNode within entities.


service/src/main/java/org/jboss/sbomer/service/feature/sbom/k8s/reconciler/OperationController.java
```diff
@@ -303,15 +299,16 @@ private UpdateControl<GenerationRequest> reconcileOperationInitialized(
      * @param secondaryResources
      * @param generationRequest
      *
-     * @return
+     * @returnw
```
vibe13, line 302: typo

service/src/main/resources/db-update/00012.sql
```diff
@@ -0,0 +1,71 @@
...
+    subquery
+WHERE
+    sbom_generation_request.id = subquery.id;
+
+INSERT INTO
+    db_version (version, creation_time)
+VALUES
+    ('00012', now ());
+
+COMMIT;
```
vibe13, line 71: Aren't there any analysis or container operations to be converted also?

vibe13 (approved):
I really like the simplification of all the code around the configs. It was ugly and difficult to maintain, thanks for this massive refactoring, having new configs will be so easy now. 
Left just 2 very minor comments for typos, and I guess adding the conversion also for analysis and container operations in the SQL script would not harm even if there are none. Thanks @goldmann !!!

service/src/main/resources/db-update/00012.sql
goldmann: Good catch, I was checking only stage database. For completeness I'll add missing types!

goldmann: This should be now covered:

```
select distinct type from sbom_generation_request;
      type      
----------------
 BUILD
 CONTAINERIMAGE
 OPERATION
(3 rows)
```

PR #593, opened 2024-07-16
UMB and S3 fixes

goldmann:
This PR contains fixes for two things:

1. UMB notifications are send only after manifests are persisted in the database which makes it possible to handle case where failure in sending messages could prevent from updating the generation status
2. Storing in S3 is enabled for failed generations as well.

https://issues.redhat.com/browse/SBOMER-102
https://issues.redhat.com/browse/SBOMER-99


vibe13 (approved):
Makes sense! Thanks!

PR #594, opened 2024-07-16
feat(SBOMER-91): add RPM content support for container images

goldmann:
Adds support for adding RPM components to the manifest. If the rpms flag is set to true, RPMs will be preserved in the manifest. Otherwise (the default option) RPM components will be removed.

https://issues.redhat.com/browse/SBOMER-91


vibe13 (approved):
Super and neat!!

PR #630, opened 2024-08-07
Only set POSIX file permissions if supported

dwalluck:
This seems to indicate that 0700 is already the default https://nvd.nist.gov/vuln/detail/CVE-2020-8908/change-record?changeRecordedOn=02/16/2021T18:15:13.103-0500 but I wouldn't believe it without checking the code.

In any case, only set POSIX perms if supported.

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/provider/BuildFinderConfigProvider.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/provider/BuildFinderConfigProvider.java
@@ -20,6 +20,7 @@ package org.jboss.sbomer.cli.feature.sbom.provider;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.attribute.PosixFilePermissions;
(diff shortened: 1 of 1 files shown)


goldmann:
I guess we don't target anything else besides Linux. Our main target is a container image.

dwalluck:
> I guess we don't target anything else besides Linux. Our main target is a container image.

Well, this small change would make it work with Windows I guess (don't know about Mac), but I was also curious that the CVE talks about the perms already being 0700, what were you seeing when you added it?

goldmann:
I was applying this change as a fix for issue discovered by Sonar.

I don't really care about Windows :)

dwalluck:
OK, the fix is still there. This was just adding a pre-check that the call won't fail on non-POSIX. 🤷

goldmann (approved):

dwalluck:
Maybe Sonar is correct and it's necessary, and this thing is wrong:

"For other Java developers, we recommend migrating to the Java 7 API `java.nio.file.Files.createTempDirectory()` which explicitly configures permissions of `700`". but I looked at the code

dwalluck:
But like I said, my intention is not removing the check. It just does: if (POSIX) then apply perms, else use default.

goldmann:
Sure, I already approved it, will merge later.

goldmann:
We will check after merge, but Sonar is currently broken, see: https://github.com/project-ncl/sbomer/pull/638#issuecomment-2275609649

PR #632, opened 2024-08-07
URL encode URLs

dwalluck:
So these URLs were either nor encoded or only partially encoded. In particular, the `':'` wasn't encoded.

### DIFF
--- a/service/src/test/java/org/jboss/sbomer/service/test/integ/feature/sbom/SBOMResourceRSQIT.java
+++ b/service/src/test/java/org/jboss/sbomer/service/test/integ/feature/sbom/SBOMResourceRSQIT.java
@@ -18,8 +18,10 @@
 package org.jboss.sbomer.service.test.integ.feature.sbom;
 
 import static io.restassured.RestAssured.given;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.net.URLEncoder;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


service/src/test/java/org/jboss/sbomer/service/test/integ/feature/sbom/SBOMResourceRSQIT.java
```diff
@@ -162,10 +166,7 @@ void testFetchSbomByPurlNotFound() {
                         .statusCode(404)
                         .body("errorId", CoreMatchers.isA(String.class))
                         .body("error", CoreMatchers.is("Not Found"))
-                        .body(
-                                "message",
-                                CoreMatchers.is(
-                                        "SBOM with purl = 'pkg:maven/org.apache.logging.log4j/log4j@2.19.0.redhat-00001?type=pom' couldn't be found"))
+                        .body("message", CoreMatchers.is("SBOM with purl = '" + PURL + "' couldn't be found"))
```
goldmann, line 169: I wouldn't change the expectation to be a variable. It's better to have strings, even if we are repeating these.

goldmann:
Proper encoding should be done automatically, when using the `pathParam`, from what I could figure out.

dwalluck:
> Proper encoding should be done automatically, when using the `pathParam`, from what I could figure out.

To be clear, although `PURL` is a URL, the whole thing is passed as a path, right? How it was is legal as a URL, but if it's being sent as an argument, then the whole thing should have been encoded.

I am trying to see if I can get a reproducer for what I saw.

goldmann:
Yah, I'm curious what you've seen. It was working fine for me locally and in CI for some time already.

Yes, the purl is part of the URL and it is encoded automatically by RestAssured. The `pathParam()` function is designed to handle exactly this case, I think.

dwalluck:
The only thing I have read so far is that maybe if it sees an absolute URL format, it only encodes part of it.

What drew my attention to the existing code was that the query part of the URL was encoded, but the rest of it was not. I am not sure how it doesn't doubly encode something that's already partly encoded, or if it's seeing it as an "absolute URL" and skipping the encoding for part of the string, which didn't seem correct to me. The whole thing is a path, so the whole thing should be encoded...

I guess if the investigation bares no fruit, we could close this.

dwalluck:
OK, looking at the message now (`"SBOM with purl = '" + purl + "' couldn't be found"`), the `purl` var is not decoded on the way out. Should it be?

goldmann:
I think you are trying to fix a problem that doesn't exist :)

service/src/test/java/org/jboss/sbomer/service/test/integ/feature/sbom/SBOMResourceRSQIT.java
```diff
@@ -162,10 +166,7 @@ void testFetchSbomByPurlNotFound() {
                         .statusCode(404)
                         .body("errorId", CoreMatchers.isA(String.class))
                         .body("error", CoreMatchers.is("Not Found"))
-                        .body(
-                                "message",
-                                CoreMatchers.is(
-                                        "SBOM with purl = 'pkg:maven/org.apache.logging.log4j/log4j@2.19.0.redhat-00001?type=pom' couldn't be found"))
+                        .body("message", CoreMatchers.is("SBOM with purl = 'pkg%3Amaven%2Forg.apache.logging.log4j%2Flog4j%402.19.0.redhat-00001%3Ftype%3Dpom' couldn't be found"))
```
goldmann, line 169: This expectation is wrong. What SBOMer returns (and should return) is exactly what it was before the proposed change. I wouldn't change this.

goldmann:
I'm going to close this PR. See the comment in the code about the expectation.

PR #633, opened 2024-08-07
Convert paths to strings rather than hardcoding the path

dwalluck:
This fixes issues with paths and also avoid the `assertEquals` with lists.

### DIFF
--- a/cli/src/test/java/org/jboss/sbomer/cli/test/unit/generate/MavenDominoGeneratorTest.java
+++ b/cli/src/test/java/org/jboss/sbomer/cli/test/unit/generate/MavenDominoGeneratorTest.java
@@ -17,8 +17,11 @@
  */
 package org.jboss.sbomer.cli.test.unit.generate;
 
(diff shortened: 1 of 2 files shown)


goldmann:
Similarly to my comment in https://github.com/project-ncl/sbomer/pull/632, I would leave expectations to be strings.

goldmann:
I don't think it was broken before, was it? If not, I wouldn't touch it, because it decreases readability of the tests. Again, I prefer having plain strings in assertions. We do target POSIX-only.

dwalluck:
Yes, technically the current code would only fail on Windows since while Java allows `/` to create paths just fine, when you print them they are going to print with `\`, e.g., `Path.of("foo/bar").toString() == "foo\bar"`;

goldmann:
Windows is not our (dev or target) platform, I would better leave the code as it was (better readability).

goldmann:
Closing this PR, see above comment.

PR #646, opened 2024-08-08
Add script to automate the setup of sdkman and java

dwalluck:


goldmann:
Thanks for this! I would leave the instructions in the doc in addition to the note about script. The reason is that someone could already have SDKMAN! installed and they would not expect to have someone mess with their config.

goldmann (changes requested):
Thanks for this! I would leave the instructions in the doc in addition to the note about script. The reason is that someone could already have SDKMAN! installed and they would not expect to have someone mess with their config.

dwalluck:
Good point. I've now left the old section entirely as it was and just split it into two sections (automatic and manual install).

dwalluck:
I compared this script to the one for the CI, and I see that the CI has this additional step to install the internal certs.

I install haven't installed the internal certs into the new JDK yet locally as there is no mention in the dev docs. I wonder if we should add something here (maybe in a different PR).

goldmann (approved):

PR #660, opened 2024-08-16
Add some missing steps to the guides

dwalluck:


docs/modules/developer-guide/pages/development-guide.adoc
```diff
@@ -51,25 +51,89 @@ If you change under following directories you will be required to rebuild images
...
+----
+
+Follow link:https://developer.hashicorp.com/vault/install[upstream documentation] if you need more information on how to install `vault`.
+
+This step is optional and only needed to update the secrets. You can `paas-encryption-cli` by running
+
+[source,console]
+----
+
+GITLAB_HOST=gitlab.cee.redhat.com glab repo clone paas/paas-encryption-cli
```
goldmann, line 108: I would not expose this here.

goldmann (changes requested):
I have mixed feeling about this one:

1. It is Fedora-specific, I would probably refrain from it. For example I don't use Fedora. I would probably just mention tools and link to installation guides upstream.
2. I would remove any mentions of the security-related bits. Internal repository documentation can be provided by pointing (by the developers) to the `sbomer-helm` repository which contains some doc as well. This is not optimal, but at lest we don't leak how things are structured internally.
3. I would remove GitHub CLI and GitLab CLI instructions, these are very dependent on the developer preference/workflow (I don't use these).

dwalluck:
I have an idea. How about I provided a script like `./hack/install-prereqs-fedora.sh` that installs all required tools, then just make one brief mention of this and remove any direct mention from the docs.

dwalluck:
@goldmann Updated the docs. Added a script instead.

vibe13:
I would also not mention anything specific related to the scripts in sbomer-helm. If anything is missing in the internal documentation, it should be there. Even if we leaked already something, we should not add anything more. Also, if you happen to change sbomer-helm, you would need to remember to update instructions here, which is harder to track.

Wrt Fedora instructions, I would add a sentence like " you need to install these packages etc etc", and then I am ok in saying that if you have Fedora, you can run the specific script etc etc. But I don't think we list everything required?

docs/modules/developer-guide/pages/environment-setup.adoc
```diff
@@ -11,6 +11,11 @@ be repeated unless you will decide to start from scratch.
 
 Please install required tools.
 
+[TIP]
+====
+If you are running Fedora, you can run `hack/fedora-setup.sh` to install many of the required tools.
+====
+
```
goldmann, line 18: Can you remove this paragraph?

hack/fedora-setup.sh
```diff
@@ -0,0 +1,33 @@
...
+
+sudo dnf -y install https://storage.googleapis.com/minikube/releases/latest/minikube-latest.x86_64.rpm
+
+sudo dnf -y install dnf-plugins-core
+sudo dnf config-manager --add-repo https://rpm.releases.hashicorp.com/fedora/hashicorp.repo
+sudo dnf -y install vault
+
+sudo dnf -y install 'dnf-command(copr)'
+sudo dnf copr enable chmouel/tektoncd-cli
+sudo dnf -y install tektoncd-cli
```
goldmann, line 33: Can you remove this file?

goldmann (changes requested):
Hey @dwalluck, can you apply requested changes? I'll merge it afterwards.

docs/modules/developer-guide/pages/environment-setup.adoc
dwalluck: Done.

hack/fedora-setup.sh
dwalluck: Done.

PR #737, opened 2024-09-12
feat: set compute resources for task runs

dwalluck:


goldmann:
So, I checked this and it looks like it is not necessary to set these values. We are setting these requirements in the Task definition (sbomer-helm repository) and the overall resource requirement is computed based on this according to https://tekton.dev/docs/pipelines/compute-resources/. This is good enough for now. We may need to tweak it per TaskRun in the future if we find that these resources are too high and it would decrease the throughput of manifest generation in case of big workloads.

So, I'll close this PR for now.

PR #776, opened 2024-09-27
fix(SBOMER-166): do not set product coordinates to not send UMB message

goldmann:
The problem is that the main component (metadata/component) does not have the build information attached, because the artifact cannot be found in PNC and thus the information about the build system wasn't added to external references in the main component.

This comimt stops setting the product coordinates for this specific e2e, but this is just a workaround. The proper fix will be implemented later.

Related: https://issues.redhat.com/browse/SBOMER-26
Fixes: https://issues.redhat.com/browse/SBOMER-166


vibe13:
I think this makes sense because we are propagating something that we know it won't work. The fix as you say would be to take a look at Gradle components which have multimodule (this is what I think was the outstanding issue with Gradle builds with the gradle-cyclonedx-plugin). So I am fine with this temporary solution until we fix the manifesting.

PR #782, opened 2024-09-28
fix(SBOMER-167): do not create new requests if we already handle it

goldmann:
Because we only ACK the message when we successfully create a `GenerationRequest` CR and sync it with the DB we probably can use this fact to check whether a `SbomGenerationRequest` entity already exists for a given PNC build.

I think it is safe to assume that if it exists, it was created by processing another message. It could help with processing duplicate UMB messages, because we have a single active consumer and we process next message only when we ACK or reject previous message.

https://issues.redhat.com/browse/SBOMER-167

WDYT @vibe13?


goldmann:
This was marked a draft until we get some pair of eyes :)

vibe13:
@goldmann Hi Marek yes I think it is a fair assumption to check in the database. This change only affects the code related to UMB messages, so we can still ask via CLI / REST the generation of the same build multiple times (usually when we change the manifesting implementation and need a newer manifest), so I am ok with this! Thanks!

vibe13 (approved):

goldmann:
Thanks! Let's give it a shot!

PR #789, opened 2024-10-01
fix(SBOMER-171):  bugfixing operation handling

goldmann:
This PR fixes some inconsistencies in operation handling. Operation schema was updated, because it was wrong according to what we generate internally. Now it is allowed to not provide product config. Handling of empty config was fixed as well.


vibe13 (approved):
Looks good, thanks!

PR #795, opened 2024-10-06
[SBOMER-138] - Create a listener to Errata Tool events

vibe13:
@goldmann Initial integration work to see how the integration with Errata API works. As we understand which data we need to fetch and which APIs, more will be added.

### DIFF
--- a/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/CustomInstantDeserializer.java
+++ b/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/CustomInstantDeserializer.java
@@ -1,0 +1,45 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.sbomer.core.features.sbom.utils;
+
+import java.io.IOException;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.TemporalAccessor;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+
+public class CustomInstantDeserializer extends JsonDeserializer<Instant> {
+
+    /*
+     * This is to convert dates like "2024-10-01 08:03:25 UTC" into a java.time.Instant
+     */
+    @Override
+    public Instant deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        String date = p.getText().trim();
+        // Remove the "UTC" and parse
+        if (date.endsWith(" UTC")) {
+            date = date.replace(" UTC", "Z");
+        }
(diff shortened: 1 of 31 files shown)


service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/ErrataMessageHelper.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jboss.sbomer.service.feature.sbom.features.umb.consumer.model.ErrataStatusChangeMessageBody;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+
+public class ErrataMessageHelper {
+
+    static ObjectMapper customJsonObjectMapper = ObjectMapperProvider.json()
+            .registerModule(new SimpleModule().addDeserializer(Instant.class, new CustomInstantDeserializer()));
```
goldmann, line 34: I fear this changes the JSON mapper we use everywhere. I don't think this was the intention, right?

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/CustomInstantDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+public class CustomInstantDeserializer extends JsonDeserializer<Instant> {
+
+    /*
+     * This is to convert dates like "2024-10-01 08:03:25 UTC" into a java.time.Instant
+     */
+    @Override
+    public Instant deserialize(JsonParser p, DeserializationContext ctxt) throws IOException, JsonProcessingException {
+        String date = p.getText().trim();
+        // Remove the "UTC" and parse
+        if (date.endsWith(" UTC")) {
```
goldmann, line 39: Does this mean that Errata returns dates in different formats?! At least the in the docs they say it's https://en.wikipedia.org/wiki/ISO_8601.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/features/umb/consumer/AmqpMessageConsumer.java
```diff
@@ -71,18 +79,54 @@ public void init(@Observes StartupEvent ev) {
 
     @Incoming("errata")
     @Blocking(ordered = false, value = "errata-processor-pool")
-    public CompletionStage<Void> processErrata(Message<String> message) {
+    public CompletionStage<Void> processErrata(Message<byte[]> message) {
+        errataReceivedMessages.incrementAndGet();
+
         log.debug("Received new Errata tool status change notification via the AMQP consumer");
-        log.debug("Message content: {}", message.getPayload());
+        log.debug("Raw Message content: {}", message.getPayload());
```
goldmann, line 86: I think we can drop this line, it pollutes the log with unreadable content.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/Stats.java
```diff
@@ -51,7 +61,8 @@ public static class Producer {
     @Data
     @Builder(setterPrefix = "with")
     public static class Messaging {
-        Consumer consumer;
+        PncConsumer pncConsumer;
+        ErrataConsumer errataConsumer;
```
goldmann, line 65: We need to update UI to reflect this change.

goldmann (changes requested):
Good stuff!

Small comment here and there. Plus, if you can -- please update commit messages to follow conventional commits standard.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/ErrataMessageHelper.java
vibe13: Right, I will use it only here instead.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/CustomInstantDeserializer.java
vibe13: Errata is fine, it's UMB messages which are not. The dates are in format "2024-10-07 07:24:42 UTC" which use a space instead of a "T" to separate the date and time, and use "UTC" instead of "Z" to indicate that the time is in UTC.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/Stats.java
vibe13: Ah right! Thanks, updated!

goldmann (approved):

PR #820, opened 2024-10-15
fix(SBOMER-198): Use string ids for UMBMessage table

vibe13:


service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/UMBMessage.java
```diff
@@ -115,4 +120,26 @@ public static long countPncReceivedMessages() {
...
+    public static UMBMessage createNew(UMBConsumer consumer) {
+        return UMBMessage.builder()
+                .withId(RandomStringIdGenerator.generate())
+                .withConsumer(consumer)
+                .withReceivalTime(Instant.now())
+                .withStatus(UMBMessageStatus.NONE)
+                .build();
+    }
+
+    public static UMBMessage ackAndSave(UMBMessage message) {
```
goldmann, line 133: Wouldn't it make sense to not make it a static method? I think it would be cleaner. Don't we require `@Transactional` here?

vibe13: I have changed my mind a couple of times already on the static - non static method. can do the non static one.  wrt @transactional : I forgot one in the parent method but yes makes sense to add it here as well. fixing now

goldmann (approved):

PR #834, opened 2024-10-21
fix(SBOMER-205): fix the retrieval of Koji configuration

vibe13:


goldmann:
Hmm, I would probably use something like this:

```java
@ConfigProperty(name = "sbomer.koji.hub-url")
Optional<String> kojiHubUrl;
```

and update Helm chart to provide the URL and not the host. This would make it probably cleaner. WDYT, @vibe13?

goldmann (approved):

PR #848, opened 2024-10-23
Container image metadata component cleanup

goldmann:
The main component representing the image should be located in the root components list (probably first one). It should contain all properties, build information, etc.

The metadata component should be a bare representation of the main component containing product release metadata, in case it is a release manifest.


vibe13 (approved):
Looks good to me!

PR #865, opened 2024-10-29
refactor: Remove Swagger annotations

vibe13:


goldmann:
I think I would just remove these, we can always get them back, if needed from the gist history.

goldmann (approved):

PR #875, opened 2024-11-06
[SBOMER-219] - Store trigger information

vibe13:


vibe13:
@goldmann I need to:
* add SQL scrips (to update the model and migrate old UMBMessages to new RequestEvent)
* double check UI
* add DTOs? Do we want to expose request events?
* maybe more

goldmann (dismissed):
OMG, this is huge, good job!

goldmann:
Related to the question of exposing this data -- I would say it;s not necessary now. Let's store it first and we will think about the proper way to present it later.

vibe13:
> Related to the question of exposing this data -- I would say it;s not necessary now. Let's store it first and we will think about the proper way to present it later.

Yep, let's first get them and then we can think what to do with such information later. Thanks!

goldmann (dismissed):

vibe13:
@goldmann I have updated the PR mainly to rebase on top of latest fixes on main, and I have implemented the check for existing messages in the DB with the new format. I have also refactored the RequestEventRepository so we don't have to duplicate the queries for H2 in AlternativeRequestEventRepository, now it's handled automatically.

goldmann (approved):

goldmann:
Let's do it!

PR #876, opened 2024-11-06
test: update e2e tests after API refactoring

goldmann:
All e2e tests had to be updated, because latest API version returns array of generations in response.

Reading of config files was adjusted because we do not default to the PncBuildConfig.

One image test was removed, because it's the same as the other one (we do not allow to provide product - errata - config anymore).


goldmann:
This is a step to fix e2e test, but this is also an improvement in handling serialization. I think I may need to do another round of e2e fixes after this is merged, but these should be very light.

vibe13 (approved):
Nice improvements on the validation of configs!

PR #880, opened 2024-11-08
Fix startup and duplicated manifests

vibe13:


goldmann (dismissed):
Hah!

```
2024-11-08T11:12:25.3255521Z [WARNING] Missing header in: /home/runner/work/sbomer/sbomer/service/src/main/java/org/jboss/sbomer/service/db/DatabaseReadinessCheck.java
```

PR #906, opened 2024-11-14
feat(SBOMER-224): adjust syft image TaskRun compute resources

goldmann:
It does this based on the configuration and it can be overridden by an env variable.

Currently covering only container image generations, will be extended to other types if it proves to be working properly.


vibe13 (approved):
Nice!

PR #917, opened 2024-11-16
fix: make the JVM settings for generator static

goldmann:
These settings control only the SBOMer CLI that launches the generator. Generator itself has own set of resource limits.

Bump memory settings for Domino and CycloneDX plugins.


goldmann:
My previous PR was actually wrong. The JVM settings should be pretty much same for all generations, what matters are the generator JVM settings. This fixes it (although these are still hardcoded, but at least more meaningful!).

PR #919, opened 2024-11-17
[SBOMER-214] - Errata integration for text-only advisories

vibe13:
This addresses also https://issues.redhat.com/browse/SBOMER-234


service/src/main/resources/schemas/appsvc-metadata.schema.json
```diff
@@ -0,0 +1,156 @@
...
+        "properties": {
+          "type": {
+            "const": "pnc-build"
+          },
+          "buildId": {
+            "description": "PNC build identifier",
+            "type": "string"
+          }
+        },
+        "required": ["buildId", "type"],
```
goldmann, line 101: Could you please update the schema in fix-mappings/schema to match this?

service/src/main/java/org/jboss/sbomer/service/feature/sbom/service/AdvisoryService.java
```diff
@@ -167,6 +243,61 @@ private Collection<SbomGenerationRequest> handleStandardQEAdvisory(RequestEvent
...
+            }
+        }
+
+        return generations;
+    }
+
+    /**
+     * Helper method to create RequestConfig based on the deliverable type.
+     */
+    private RequestConfig createRequestConfig(String type, JsonNode deliverableEntry) {
```
goldmann, line 283: It's ok for now, but I would like to move this https://github.com/project-ncl/sbomer/blob/e560e9462a0768b9811efcdc785c1514af0bcd27/core/src/main/java/org/jboss/sbomer/core/features/sbom/config/Config.java#L121-L221 to a utility class and use instead of what we have here.

It provides much more detailed information in case of failures.

goldmann (approved):

service/src/main/resources/schemas/appsvc-metadata.schema.json
vibe13: sure was on my list :-)

PR #927, opened 2024-11-18
fix(SBOMER-242): fix log message to use correct identifier

goldmann:


vibe13 (approved):
I am finding a hard time understanding how this could even happen, but ok!

PR #930, opened 2024-11-18
fix(SBOMER-244): use different subscription IDs across AZ to handle load in parallel

goldmann:
Different subscription IDs should make clients receive messages in parallel.

### DIFF
--- a/service/src/main/resources/application.yaml
+++ b/service/src/main/resources/application.yaml
@@ -29,7 +29,7 @@ mp:
     incoming:
       errata:
         port: ${UMB_BROKER_PORT}
-        address: ${SBOMER_CONSUMER_ERRATA_TOPIC}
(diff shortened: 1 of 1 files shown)


goldmann:
@vibe13 Is my assumption correct? I think the load should be spread across consumers with different subscription ID. It will be different, because the `SBOMER_DEPLOYMENT_ZONE` will be different in our case.

vibe13:
@goldmann Yes your assumption is correct! 
If a service account is named "my-umb-client" and they have been granted read access to "VirtualTopic.foo", they will be able to access a consumer queue named "Consumer.my-umb-client.foo-subscription.VirtualTopic.foo" ($SERVICE_ACCOUNT_NAME = "my-umb-client", $SUBSCRIPTION_ID = foo-subscription). **If "my-umb-client" connects another consumer to the same queue name, only one of the consumers will get a message that was published to the original topic. If they needed another copy of the message for this other consumer they would create another consumer queue with a different $SUBSCRIPTION_ID.**  For example, "Consumer.my-umb-client.bar-subscription.VirtualTopic.foo".  This is actually one of the advantages of virtual topics over regular topics, consumers have more flexibility in how they receive and process messages.

goldmann:
So, this should actually make use of both clusters, then. Ready for review!

vibe13 (approved):
Let's do it!

PR #952, opened 2024-11-20
fix(SBOMER-236): Dependency reference does not exist in the BOM

vibe13:


goldmann:
I'm reviewing it, but I would like to get to the bottom of the issue, give me some time.

vibe13:
As discussed let's close this one

PR #955, opened 2024-11-20
chore: simplify setting of topic addresses

goldmann:


goldmann:
This is used only for dev, for stage and prod we set it using other env vars.

vibe13 (approved):

PR #957, opened 2024-11-21
chore: generator upgrades: maven-cdx 2.9.0, node-cdx 1.19.3, gradle-c…

goldmann:
…dx: 1.10.0


vibe13 (approved):
Looks good, we are still defaulting to 1.4 spec on all of them so we are consistent.

goldmann:
@vibe13 I just pushed an update to ensure we specify the schema version everywhere. Gradle does it in `images/sbomer-generator/cyclonedx-init.gradle`.

vibe13 (approved):

PR #959, opened 2024-11-21
fix(SBOMER-200): Fix possible NPE in findBuildByRPM

janinko:


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/service/KojiService.java
```diff
@@ -217,6 +217,12 @@ public KojiBuildInfo findBuildByRPM(String nvra) throws KojiClientException {
             return null;
         }
         KojiRpmInfo rpmInfo = rpm.get(0);
```
goldmann, line 219: Can it happen that the `rpm` is `null` as well?

goldmann (approved):

PR #970, opened 2024-11-22
ci: improve speed of documentation building

goldmann:
This takes the generation of documentation from 7 min to 2.


vibe13 (approved):
wow! great

PR #993, opened 2024-11-28
feat(SBOMER-142): Enhance handling of bundled JavaScript content

janinko:


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
```diff
@@ -266,9 +275,102 @@ public Bom process(Bom bom) {
...
+
+    private void addMainComponentDependencies(Bom bom, Set<String> dependencies) {
+        String mainComponentRef = bom.getMetadata().getComponent().getBomRef();
+
+        Dependency mainComponentDependencies = bom.getDependencies().stream()
+            .filter(d -> mainComponentRef.equals(d.getRef()))
+            .findFirst()
+            .orElseGet(() -> addNewDependency(bom, mainComponentRef));
+
+        dependencies.stream().map(SbomUtils::createDependency).forEach(mainComponentDependencies::addDependency);
```
vibe13, line 356: From what I understand, this is creating a new dependency and adding it to the mainComponentDependencies (so it ends up in the dependsOn[]), but should it also add the new dependency to the bom? So that each dependency is also listed singularly (with an empty dependsOn[]) ?
I guess I want to say, should we also have a `bom.addDependency(newDep)` for each newly created dependency?

vibe13 (commented):
I put a comment because I think we are missing something, but otherwise looks good to me!

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
```diff
@@ -266,9 +275,102 @@ public Bom process(Bom bom) {
...
+                log.warn("Unable to parse artifact purl '{}'.", artifact.getPurl());
+                continue;
+            }
+            if (listedPurls.contains(coordinates)) {
+                continue;
+            }
+            Component newComponent = createComponent(artifact, Component.Scope.REQUIRED, Component.Type.LIBRARY, buildType);
+            setArtifactMetadata(newComponent, artifact, pncService.getApiUrl());
+            setPncBuildMetadata(newComponent, artifact.getBuild(), pncService.getApiUrl());
+            bom.addComponent(newComponent);
```
goldmann, line 342: @vibe13 I think the component is added here. And it looks that the tests cover this.

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
goldmann: I think you missed the line I highlighted here: https://github.com/project-ncl/sbomer/pull/993/files#r1866566103

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
```diff
@@ -266,9 +275,102 @@ public Bom process(Bom bom) {
...
+            return;
+        }
+        Component component = bom.getMetadata().getComponent();
+        List<ExternalReference> externalReferences = getExternalReferences(component, ExternalReference.Type.BUILD_SYSTEM, SBOM_RED_HAT_PNC_BUILD_ID);
+        if (externalReferences.isEmpty()) {
+            // Not a PNC build
+            return;
+        }
+        if (externalReferences.size() > 1) {
+            log.warn("Component {} hase more then one {}/{} external reference", component.getPurl(), ExternalReference.Type.BUILD_SYSTEM, SBOM_RED_HAT_PNC_BUILD_ID);
```
goldmann, line 295:
```suggestion
            log.warn("Component {} hase more than one {}/{} external reference", component.getPurl(), ExternalReference.Type.BUILD_SYSTEM, SBOM_RED_HAT_PNC_BUILD_ID);
```

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
```diff
@@ -266,9 +275,102 @@ public Bom process(Bom bom) {
         // If there are any purl relcoations, process these.
         purlRelocations.forEach((oldPurl, newPurl) -> updatePurl(bom, oldPurl, newPurl));
 
+        addMissingNpmDependencies(bom);
```
goldmann, line 278: We do process only the main component? Shouldn't it be performed as part of the `processComponent` method to cover all potential cases?

vibe13: You mean to get all the transitive NPM dependencies for all the dependent components?  Is the Maven CycloneDX plugin traversing all the transitive dependency tree to create the component list? I don't remember if that is the case. If yes then I agree we should probably move `addMissingNpmDependencies` to all the components and not only the main one. Otherwise I believe it's fine?

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
vibe13: I guess what I meant is that I cannot see where each new dependency is added to the bom. 
For every NPM artifact:
* I see that a new Component is added to the BOM
* I see that a new Dependency is created and that dependency is added to the main component dependency list
* I do not see that the new Dependency is also added to the BOM. Something like:
 
```
        dependencies.stream().map(SbomUtils::createDependency).forEach(d -> {
            mainComponentDependencies.addDependency(d);
            bom.addDependency(d);
        });
```
The original dependency section of the BOM for the main component is:
```
    {
      "ref": "pkg:maven/org.keycloak/keycloak-parent@24.0.6.redhat-00001?type=pom",
      "dependsOn": [
        "pkg:maven/org.keycloak/keycloak-common@24.0.6.redhat-00001?type=jar"
      ]
    }
```

With this PR original code, this section changed in the processed BOM:
```
    {
      "ref": "pkg:maven/org.keycloak/keycloak-parent@24.0.6.redhat-00001?type=pom",
      "dependsOn": [
        "pkg:maven/org.keycloak/keycloak-common@24.0.6.redhat-00001?type=jar",
        "pkg:npm/%40redhat/kogito-tooling-keyboard-shortcuts@0.9.0-2",
        "pkg:npm/once@1.4.0"
      ]
    }
```
and nothing else. I would have expected also 2 more dependencies. With my suggestion ^, these are the changes:
```
    {
      "ref": "pkg:maven/org.keycloak/keycloak-parent@24.0.6.redhat-00001?type=pom",
      "dependsOn": [
        "pkg:maven/org.keycloak/keycloak-common@24.0.6.redhat-00001?type=jar",
        "pkg:npm/%40redhat/kogito-tooling-keyboard-shortcuts@0.9.0-2",
        "pkg:npm/once@1.4.0"
      ]
    },
    {
      "ref": "pkg:npm/%40redhat/kogito-tooling-keyboard-shortcuts@0.9.0-2",
      "dependsOn": []
    },
    {
      "ref": "pkg:npm/once@1.4.0",
      "dependsOn": []
    }
```
which I think are more correct for the hierarchy consistency.

vibe13 (changes requested):

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
goldmann: I think with my previous comment I wanted to touch on two things:

1. We probably should process for additional NPMs only components which are of type other than RPM and container images.
2. According to the code `addMissingNpmDependencies(Bom bom)` method is covering only the component located in the under the `.metadata.component`. Why aren't we looking at every component in `.components`? I imagine that the main use case that trigerred this RFE is the manifestation of a ZIP deliverable. Why would we want to not do it for any (Java?) builds?

What would the processed manifest look if we would scan the whole component list? I think for a ZIP it should be the same, correct?

janinko:
Yeah, I think I've seen the dependencies to be like that. I'll add it.

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
janinko: @vibe13 I guess you are better at answering this. I was going according to what I understood from your "1 PNC build == 1 SBOM"

janinko: Ok, now I've read the comments. I was doing it with the understanding this is for the analysis of single PNC build. 

I do not fully understand what is the problem with the manifestation of the zip deliverable and how the deliverable analysis is turned into manifest.

vibe13: I think that for the single PNC builds we are fine with this PR. 
For the ZIP manifestation (see `CycloneDxGenerateOperationCommand`) we will need to hook in this logic to retrieve the NPM dependencies for all the components matched with PNC builds. I think I am fine with another PR for that! :-)

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java
```diff
@@ -266,9 +275,105 @@ public Bom process(Bom bom) {
...
+            return;
+        }
+        Component component = bom.getMetadata().getComponent();
+        List<ExternalReference> externalReferences = getExternalReferences(component, ExternalReference.Type.BUILD_SYSTEM, SBOM_RED_HAT_PNC_BUILD_ID);
+        if (externalReferences.isEmpty()) {
+            // Not a PNC build
+            return;
+        }
+        if (externalReferences.size() > 1) {
+            log.warn("Component {} hase more than one {}/{} external reference", component.getPurl(), ExternalReference.Type.BUILD_SYSTEM, SBOM_RED_HAT_PNC_BUILD_ID);
```
vibe13, line 295: `hase` :-)

vibe13 (approved):

PR #994, opened 2024-11-28
Upgrade domino to 0.0.117 that generates CycloneDX SBOMs using the 1.6 schema by default

aloubyansky:
This version generates SBOMs using the CycloneDX schema version 1.6 by default. But also introduces the `--cdx-schema-version` argument to request a different schema version.

### DIFF
--- a/core/src/main/resources/META-INF/sbomer-config.yaml
+++ b/core/src/main/resources/META-INF/sbomer-config.yaml
@@ -22,7 +22,7 @@ sbomer:
     default-generator: maven-cyclonedx
     generators:
       maven-domino:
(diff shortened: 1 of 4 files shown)


goldmann:
@aloubyansky Thanks for the PR! Can you update the commit message to follow https://www.conventionalcommits.org/en/v1.0.0/ Something like: `chore: upgrade domino to 0.0.117 that generates CycloneDX SBOMs using the 1.6 schema by default` would do it.

aloubyansky:
Done

vibe13 (approved):

PR #997, opened 2024-11-29
test: disable rpm generator test for now

goldmann:


vibe13:
Is this still due to SBOMer storing manifests with 1.4 spec?

vibe13 (approved):

PR #1010, opened 2024-11-30
feat(SBOMER-263): Handle big advisories better

vibe13:


service/src/main/java/org/jboss/sbomer/service/feature/sbom/service/AdvisoryService.java
```diff
@@ -303,20 +305,23 @@ protected Collection<SbomGenerationRequest> processDockerBuilds(
...
+                .stream()
+                .flatMap(List::stream)
+                .map(BuildItem::getId)
+                .collect(Collectors.toList());
+
+        Map<Long, String> imageNamesFromBuilds = getImageNamesFromBuilds(buildIds);
+        imageNamesFromBuilds.forEach((buildId, imageName) -> {
+            log.debug("Retrieved imageName '{}' for buildId {}", imageName, buildId);
+            if (imageName != null) {
+                config.setImage(imageName);
```
goldmann, line 319: Shouldn't we create in the loop the `config`  object? My worry is that if we reuse the object, we may interfere with the persistence layer and (maybe) accidentally sync the wrong thing to the database when the `SbomGenerationRequest.sync` is called.

vibe13: Mmmm ok, maybe. I moved it in the inner loop. Thanks!

goldmann (approved):

PR #1011, opened 2024-12-01
feat(SBOMER-192): add comments to Errata upon completion request events

vibe13:


service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/event/CommentAdvisoryOnRelevantEventsListener.java
```diff
@@ -0,0 +1,360 @@
...
+            // Text-only advisories with "manifest" notes content fall in this category.
+            handleManualManifestationAdvisory(errata, config);
+        }
+    }
+
+    private void handleAutomatedManifestationAdvisory(
+            RequestEventStatusUpdateEvent event,
+            ErrataAdvisoryRequestConfig config) {
+        List<V1Beta1RequestRecord> requestRecords = sbomService
+                .searchAggregatedResultsNatively("id=" + event.getRequestEventId());
```
goldmann, line 105: Do we specify explicit order for this query? For example last are first in the result list.

goldmann (approved):
This is superb! Great work!

service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/event/CommentAdvisoryOnRelevantEventsListener.java
vibe13: Yes the sorting is done here: https://github.com/project-ncl/sbomer/blob/a97133c8f3eb4624d285a3ebce1a4228d57f012d/service/src/main/java/org/jboss/sbomer/service/feature/sbom/service/RequestEventRepository.java#L436

So it's for receivalTime in descending order (we will get last requests first)

PR #1012, opened 2024-12-01
feat(SBOMER-225): Adjust bom-ref for container image manifests

vibe13:


goldmann:
I like to think of `bom-ref` as an identifier in the database sense, a random string. I know you don't agree with it, so we can use purl, but it needs to be consistent if we decide to do it this way.

vibe13:
@goldmann I will get back to this today

vibe13:
@goldmann ok I have added the same behavior to the DefaultProcessor, and additional checks to avoid duplication of bom-refs (happened with Syft)

goldmann (approved):

PR #1026, opened 2024-12-06
feat(SBOMER-271): CycloneDX 1.6: SBOMer can't not parse .dependencies…

vibe13:
…[].provides

### DIFF
--- a/core/src/main/java/org/cyclonedx/model/Dependency.java
+++ b/core/src/main/java/org/cyclonedx/model/Dependency.java
@@ -1,0 +1,121 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2023 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


goldmann (approved):

core/src/main/java/org/cyclonedx/model/Dependency.java
```diff
@@ -0,0 +1,121 @@
...
+public class Dependency extends BomReference {
+
+    @JsonProperty("dependsOn")
+    @JacksonXmlElementWrapper(useWrapping = false)
+    @JacksonXmlProperty(localName = "dependency")
+    private List<Dependency> dependencies;
+
+    @JsonProperty("provides")
+    @JacksonXmlElementWrapper(useWrapping = false)
+    @JacksonXmlProperty(localName = "dependency")
```
goldmann, line 61: I'm not sure if this is correct.

goldmann: https://cyclonedx.org/docs/1.6/xml/#type_dependencyType

vibe13: I have tried to get it from the spec https://github.com/CycloneDX/cyclonedx-core-java/blob/master/src/main/resources/bom-1.6.schema.json#L1847 and it looks like `dependsOn` and `provides` have the same definition...but did not find any pointers for the xml schema.

goldmann: Yah, for our use case it doesn't matter, but when submitting upstream, proper XML support may be necessary. We can always ask them to help us understand how it should be done for XML.

PR #1057, opened 2024-12-15
chore: split and import certs

goldmann:


.gitlab-ci.yml
```diff
@@ -68,15 +68,17 @@ build:
...
     - echo "sdkman_auto_answer=true" > $HOME/.sdkman/etc/config
     - source "$HOME/.sdkman/bin/sdkman-init.sh"
     - sdk install java 17.0.12-tem
-    - keytool -import -trustcacerts -alias Current-IT-Root-CAs -file certs/Current-IT-Root-CAs.pem -keystore "$HOME/.sdkman/candidates/java/17.0.12-tem/lib/security/cacerts" -noprompt -storepass changeit
+    - >
+      mkdir certs
+      echo "$IT_ROOT_CA_2022" > certs/2022-IT-Root-CA.pem
+      echo "$IT_ROOT_CA_2015" > certs/2015-IT-Root-CA.pem
+
+      keytool -import -trustcacerts -alias 2022-IT-Root-CA -file 2022-IT-Root-CA.pem -cacerts -noprompt -storepass changeit
```
vibe13, line 80: don't we need the path here?

keytool -import -trustcacerts -alias 2022-IT-Root-CA -file certs/2022-IT-Root-CA.pem -cacerts -noprompt -storepass changeit

same below

goldmann: In newer JDKs (11+) we don't need to specify path to keystore. The `-cacerts` takes care of it automagically. And we know it will be on JDK 17 (or newer).

goldmann: Fixed the path to cert!

vibe13 (approved):
Let's go!!! :-D

PR #1066, opened 2024-12-18
feat(SBOMER-142): Enhance handling of bundled JavaScript content

janinko:
This will enhance all components in manifests generated from deliverables.

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/CycloneDxGenerateOperationCommand.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/CycloneDxGenerateOperationCommand.java
@@ -206,13 +206,11 @@ public class CycloneDxGenerateOperationCommand extends AbstractGenerateOperation
             // pom.xml)
             if (!purlToComponents.containsKey(artifact.getArtifact().getPurl())) {
                 KojiBuild brewBuild = null;
-                BuildType buildType = null;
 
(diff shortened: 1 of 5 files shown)


vibe13 (commented):
@janinko Hi! The changes look good, however I have a question on the code which popped up now but was already merged.
In https://github.com/project-ncl/sbomer/blob/e3db2746a743da91839519195b5b00e1291e4d6f/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/processor/DefaultProcessor.java#L357, you are creating a new component , from an artifact which has type NPM (since you want all the NPM dependencies), but using the buildType of the parent build which is of type != NPM, e.g. Maven. How does https://github.com/project-ncl/sbomer/blob/main/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java#L318 resolve the coordinates to assign to the new component? In https://github.com/project-ncl/sbomer/blob/main/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java#L164, you have a MVN buld type, but the coordinates are of type NPM, so how is it possible? Am I overlooking something? 
Thanks!

janinko:
@vibe13 you are right it was not working properly. I've fixed it in tthe new commit and also make it use proper way of parsing the artifact data, which makes the build type unnecessary.

vibe13 (approved):
Ok this looks much better. Thanks Honza!

PR #1073, opened 2024-12-19
feat: add retries when connecting to Koji

vibe13:


service/src/main/java/org/jboss/sbomer/service/feature/sbom/service/AdvisoryService.java
```diff
@@ -531,15 +544,25 @@ private Long findErrataProductVersionIdByName(ErrataProduct product, String pVer
...
+    @Retry(maxRetries = 10, retryOn = KojiClientException.class)
+    protected KojiClientSession getKojiSession() throws KojiClientException {
+        if (kojiSession == null) {
+            kojiSession = kojiProvider.createSession();
+        }
+        return kojiSession;
+    }
+
+    // This method will be retried up to 10 times if a KojiClientException is thrown
+    @Retry(maxRetries = 10, retryOn = KojiClientException.class)
```
goldmann, line 556: What types of failures are we catching with this?

vibe13: This ones:

`Unable to fetch containers information for buildIDs: [XXX]: com.redhat.red.build.koji.KojiClientException: Connection to koji at https://YYYY/brewhub is closed. Perhaps it failed to initialize?
`
This one happened at `10:58:21,042`, and there was an error just before: 
` (Reason: {faultCode=1014, faultString=database outage})` which probably made the connection unstable.

We don't attempt any reconnection so I wanted to add more fault tolerance here.

goldmann: Right, I was more thinking if we will catch errors that we can safely expect. For example if we would catch NotFoundException (imaginary one), we would retry 10 times, but in fact this exception should be ignored. My question was about what are the types of exceptions that `KojiClientException` catches.

vibe13: That would be definitely connection issues, plus an error in the execution of the XmlRpc call:

```
private <T> T doXmlRpcAndThrow( KojiInternalCommand<T> cmd, String message, Object... params )
            throws KojiClientException
    {
        checkConnection();

        try
        {
            return cmd.execute();
        }
        catch ( XmlRpcException e )
        {
            throw new KojiClientException( "%s. Reason: %s", e, String.format( message, params ), e.getMessage() );
        }
    }
 ```

goldmann: Great, this is exactly what we need!

goldmann (approved):

PR #1076, opened 2024-12-23
chore: reduce time spent in Quarkus tests

gsmet:
This is a follow up of Marek asking for help here: https://quarkusio.zulipchat.com/#narrow/channel/187030-users/topic/.40QuarkusTest.20best.20practices.20.28multiple.20tests.2C.20slowness.29

From the first commit comment (the second one is just going a step further):

> Some background: Quarkus is restarted whenever a test resource is
> restricted to the class or different from the ones from other tests and
> also when the test profile is different.
> 
> There is a clever orderer in Quarkus that tries to order things cleverly
> to minimize the number of restarts but... ITs and Failsafe are not
> honoring the JUnit test orderer so you lose this benefit and you end up
> with Quarkus being restarted a lot more often than strictly necessary.
> 
> So the first fix here is to consider the Quarkus tests as Surefire
> tests, not failsafe tests.
> If more practical for you, they can be isolated in a specific Surefire
> execution but we should keep them executed with Surefire.
> 
> Second issue is that @WithKubernetesTestServer is restricted to a given
> test clean, meaning every time you have this annotation, the server will
> have to be restarted.
> I will start discussions to see if we should change that but in the
>  meantime I added an annotation that has a less aggressive behavior.
> 
> Unfortunately, you are also affected by a Quarkus bug, for which I will
> push a fix soon and that we will hopefully backport to the next 3.17.

Combined with https://github.com/quarkusio/quarkus/pull/45253 , I was able to reduce the time spent in `service` module testing from `4mn30` to `1m45` and it should scale a lot better if you don't multiply the combination of resources + test profiles.

There's one VERY important caveat though: the Kubernetes mock is shared with all the tests so you need to have this limitation in mind. Marek told me that was acceptable so here we are.


service/src/test/java/org/jboss/sbomer/service/test/integ/feature/sbom/k8s/reconciler/GenerationPhaseGenerationRequestReconcilerTest.java
```diff
@@ -449,7 +450,9 @@ void testFailedWithoutCleanup(@TempDir Path tempDir) throws Exception {
                 .reconcile(request, mockContext(Collections.emptySet()));
 
         assertTrue(updateControl.isNoUpdate());
-        assertNull(mockServer.getLastRequest());
+        // there is no proper way to reset the last request at the beginning of the method
+        // so let's just compare it's the same as before and no requests have been executed.
+        assertSame(lastRequestPriorToExecution, mockServer.getLastRequest());
```
gsmet, line 455: I created https://github.com/fabric8io/kubernetes-client/issues/6750 for that.

gsmet:
FYI, I'm currently releasing Quarkus 3.17.6 that will come with the Quarkus fix that should improve things for you even more with this patch. I'll include the upgrade in this PR if not merged by then.

goldmann:
Thanks for heads-up! I actually plan to merge this one now. I can do the Quarkus upgrade myself once it becomes available, or, if you are so kind, you could create a new PR. Hope this works for you.

@gsmet  Thanks for the support!

gsmet:
Works for me, I'll create a PR once 3.17.6 has been released.

PR #1077, opened 2024-12-23
chore: update to latest Quarkus micro

gsmet:
While I agree you should stay compatible with the minor requested by the Operator SDK extension, I don't think it's a good idea to stay on the micro as you are missing all the fixes we are pushing since at the moment the version used here is 3.17.0.

So my advice is: update to the latest micro of the minor branch requested for the Operator SDK.

This is important for #1076 as otherwise, you will have to wait a lot longer to get the fix. Also it's good practice in general.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,11 +53,12 @@
     <version.surefire-plugin>3.5.2</version.surefire-plugin>
 
     <!--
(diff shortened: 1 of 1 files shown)


goldmann:
Thanks a lot. I think it makes sense. We will use this approach from now on.

/cc @vibe13 @janinko

PR #1095, opened 2025-01-16
fix(SBOMER-142): Really add the NPM dependencies in Operation generation

janinko:
The operation generation actually doesn't use the Processor where the
addition was taking place previously. This moves it to
CycloneDxGenerateOperationCommand directly.

Also this implementaion doesn't query PNC multiple times for the same
build. And it will not repeat same dependnecy if it comes from multiple
builds.

+ refactor: simplify and optimize component creation for OperationCommand


vibe13 (approved):
Nice improvements Honza!

PR #1096, opened 2025-01-20
fix: add temporary build value to the valid build check error message

dwalluck:


dwalluck:
This was reported by a user who didn't realize that the build could not be temporary.

vibe13 (approved):

PR #1098, opened 2025-01-27
style: Clean up code to improve sonar low severity numbers

dwalluck:


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/service/KojiService.java
```diff
@@ -154,7 +154,7 @@ private List<KojiBuild> awaitResults(Future<List<KojiBuild>> finderTask)
             throws InterruptedException, ExecutionException {
 
         int retry = 1;
-        while (!finderTask.isDone() && (retry * 500) < MAX_BREW_WAIT_5_MIN) {
+        while (!finderTask.isDone() && (retry * 500L) < MAX_BREW_WAIT_5_MIN) {
             try {
                 retry++;
                 Thread.sleep(500);
```
dwalluck, line 160: This busy-waiting I think comes from deliverables-analyzer, we should investigate it at some point. Something doesn't smell right to me.

vibe13 (approved):

PR #1103, opened 2025-01-28
test: Additional E2E tests for skinny image manifest type

jonathanchristison:
This PR adds an additional two images to check with E2E tests, the test input json is now an array as opposed to multiple payloads in separate files.

This test would of originally failing due to alternative image manifest types (non-fat manifest) not being handled by the tekton inspect tasks, this simply tests for regression.

What I'm not sure on 

- If we actually need two additional image? 
  - These where the original generations in the bug 
  - There is also additional images that where used during testing but removed (internal url references), these fail later for a different reason
- If we only have one image to test is it worth keeping the `@ParameterizedTest`? The only advantage I can think of is if we find something similar in future and a list of images is easier


goldmann (changes requested):
@jonathanchristison Can you please format commit messages according to https://www.conventionalcommits.org/en/v1.0.0/?

goldmann (approved):

PR #1106, opened 2025-01-29
fix: Add at least one assertion to every test case

dwalluck:


service/src/test/java/org/jboss/sbomer/service/test/integ/rest/RestResourceTest.java
```diff
@@ -111,7 +111,7 @@ UmbConfig umbConfig() {
     void testListSbomsPageParams(TestableApiVersion apiVersion) {
         Mockito.when(sbomService.searchSbomRecordsByQueryPaginated(1, 20, null, null)).thenReturn(new Page<>());
         given().when()
-                .get(String.format("?pageIndex=1&pageSize=20", apiVersion.manifestsPath()))
+                .get(String.format("%s?pageIndex=1&pageSize=20", apiVersion.manifestsPath()))
```
dwalluck, line 114: This passed before even though I think the URL was incorrect?

service/src/test/java/org/jboss/sbomer/service/test/integ/feature/sbom/messaging/AmqpMessageConsumerTest.java
```diff
@@ -31,22 +33,21 @@
...
-        Message<String> message = AmqpMessageHelper.toMessage(TestResources.asString("umb/unexpected.json"), headers);
-
-        consumer.process(message);
+    void testParsingOfUnexpectedData() {
+        Map<String, Object> map = Map.of("type", "DeliverableAnalysisStateChange",
+            "timestamp", 1698076061381L,
+            "message-id", "ID:orch-86-qmrdq-33543-1697588407649-5:1:3:1:1",
+            "destination", "/topic/VirtualTopic.eng.pnc.builds");
+        JsonObject headers = new JsonObject(map);
+        assertDoesNotThrow(() -> consumer.process(AmqpMessageHelper.toMessage(TestResources.asString("umb/unexpected.json"), headers)));
```
dwalluck, line 51: OK, I didn't need to change the constructor, line 51 is the actual change.

service/src/test/java/org/jboss/sbomer/service/test/unit/feature/sbom/errata/ReleaseAdvisoryEventsListenerTest.java
```diff
@@ -729,6 +731,7 @@ void testTextOnlyReleaseErrataWithManifests() throws IOException {
                 .withReleaseGenerations(pvToGenerations)
                 .build();
         listenerTextOnlyManifests.onReleaseAdvisoryEvent(event);
+        event.getReleaseGenerations().values().forEach(request -> assertNotEquals(RequestEventStatus.FAILED, request.getRequest().getEventStatus()));
```
dwalluck, line 734: Is this correct? It seems the status is not `SUCCESS` but `IN_PROGRESS`, so we may check for that instead.

The previous code wasn't really checking anything (the method does not throw any exceptions, so it's unlikely to fail even if it doesn't work properly).

dwalluck: You have a point. Sonar complains it is a test method with no assertions. Here I was checking `event and not `listenerTextOnlyManifests`, so at least it's not duplicating any of the other assertions in the listener.

vibe13 (approved):

PR #1116, opened 2025-02-03
style: Fix more Sonar issues

dwalluck:


service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/event/release/ReleaseTextOnlyAdvisoryEventsListener.java
```diff
@@ -304,7 +305,7 @@ protected Sbom saveReleaseManifestForTextOnlyAdvisories(
             String toolVersion,
             SbomGenerationRequest releaseGeneration,
             Bom productVersionBom,
-            List<Sbom> sboms) {
+            List<Sbom> sboms) throws JsonProcessingException {
```
dwalluck, line 308: So this is an actual code change, I removed a catch somewhere and allowed the exception to propagate up. I hope that's OK.

vibe13 (approved):

PR #1131, opened 2025-02-07
style: Fix more sonar issues with core

dwalluck:


core/src/main/java/org/jboss/sbomer/core/errors/ApplicationException.java
```diff
@@ -20,21 +20,17 @@
...
     }
 
     @Override
-    public synchronized String getMessage() {
-        if (formattedMessage == null) {
-            formattedMessage = MessageFormatter.arrayFormat(super.getMessage(), params).getMessage();
-        }
-        return formattedMessage;
+    public String getMessage() {
+        return message;
```
dwalluck, line 34: Sonar complained about the field `formattedMessage`.

However, the code itself was, for some reason, initializing the formatted message inside the `getMessage()` instead of the constructor. Also, `params` could be `null` there, which would lead to a `NullPointerException`, so I guarded against that.

core/src/main/java/org/jboss/sbomer/core/config/request/ErrataAdvisoryRequestConfig.java
```diff
@@ -38,15 +38,15 @@
...
     /**
      * Advisory identifier (number or name).
      */
-    String advisoryId;
+    private String advisoryId;
 
+    @Override
+    public String getType() {
+        return TYPE_NAME;
+    }
```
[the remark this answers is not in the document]
dwalluck: But what happens if it reads in the `type` and that that is not equal to `TYPE_NAME`?

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
```diff
@@ -800,10 +802,10 @@ public static Bom fromJsonNode(JsonNode jsonNode) {
         }
     }
 
-    public static String[] computeNVRFromContainerManifest(JsonNode jsonNode) {
+    public static List<String> computeNVRFromContainerManifest(JsonNode jsonNode) {
         Bom bom = fromJsonNode(jsonNode);
         if (bom == null || bom.getComponents() == null || bom.getComponents().isEmpty()) {
-            return null;
+            return List.of();
```
dwalluck, line 808: Converted from primitive array to `List`, but it may be better to have an `NVR` type.

vibe13 (approved):

PR #1134, opened 2025-02-07
style: Fix some more sonar issues with service

dwalluck:


service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/event/release/ReleaseStandardAdvisoryEventsListener.java
```diff
@@ -629,11 +626,13 @@ protected Sbom saveReleaseManifestForDockerGeneration(
...
-                            // FIXME: Result of String.replace() is ignored
                             manifestBom.getMetadata()
                                     .getComponent()
-                                    .getDescription()
-                                    .replace(buildManifest.getRootPurl(), rebuiltPurl);
+                                    .setDescription(
+                                            manifestBom.getMetadata()
+                                                    .getComponent()
+                                                    .getDescription()
+                                                    .replace(buildManifest.getRootPurl(), rebuiltPurl));
```
dwalluck, line 635: @vibe13 I think this should fix the issue where description wasn't actually being changed.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/features/generator/AbstractController.java
```diff
@@ -211,7 +211,7 @@ protected boolean isFinished(TaskRun taskRun) {
     protected Boolean isSuccessful(TaskRun taskRun) {
         if (!isFinished(taskRun)) {
             log.trace("TaskRun '{}' still in progress", taskRun.getMetadata().getName());
-            return null;
+            return null; // FIXME: This is not really binary, but trinary state
```
dwalluck, line 214: Sonar is complaining about all these comparisons that do not use a primitive type.

service/src/main/java/org/jboss/sbomer/service/rest/criteria/predicate/CustomizedPredicateBuilderStrategy.java
```diff
@@ -108,7 +106,6 @@ public <T> Predicate createPredicate(
     }
 
     private String preprocessLikeOperatorArgument(String argument) {
-        return argument.replaceAll("_", Matcher.quoteReplacement(ESCAPE_CHAR + "_"))
-                .replaceAll("\\" + WILDCARD_CHAR, "%");
+        return argument.replace("_", ESCAPE_CHAR + "_").replaceAll("\\" + WILDCARD_CHAR, "%");
```
dwalluck, line 109: The first one is not a regex. Does the second one need to be a regex?

service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/event/AdvisoryEventUtils.java
```diff
@@ -139,19 +143,22 @@ public static Set<String> createPurls(
...
                 .map(repository -> createPurl(repository, version, includeRepositoryQualifiers))
                 .filter(Objects::nonNull)
-                .sorted(Comparator.comparingInt(String::length).reversed()) // longest first
-                .collect(Collectors.toSet());
+                .sorted(
+                        Comparator.comparingInt(String::length)
+                                .reversed() // longest first
+                                .thenComparing(Comparator.naturalOrder()))
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+        return Collections.unmodifiableSet(set);
```
dwalluck, line 154: I fixed this to use `LinkedHashSet` so that it's actually sorted.

This change also affects the test (that actually test the incorrect ordering).

I added the `.thenComparing(Comparator.naturalOrder()))` to handle strings of the same length so that we get a predictable ordering.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/RequestEvent.java
```diff
@@ -147,4 +149,35 @@ public RequestEvent save() {
...
+        RequestEvent requestEvent = (RequestEvent) o;
+        return Objects.equals(id, requestEvent.id);
+    }
+
+    @Override
+    public final int hashCode() {
+        return (this instanceof HibernateProxy)
+                ? ((HibernateProxy) this).getHibernateLazyInitializer().getPersistentClass().hashCode()
+                : getClass().hashCode();
+    }
```
dwalluck, line 182: JPA Buddy says to not use `@Data` and `@EqualsAndHashCode` with JPA entities.

https://jpa-buddy.com/blog/lombok-and-jpa-what-may-go-wrong/

service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/RequestEvent.java
```diff
@@ -147,4 +149,35 @@ public RequestEvent save() {
         return this;
     }
 
+    @Override
+    public final boolean equals(Object o) {
```
goldmann, line 153: @vibe13 I saw @dwalluck mentioning https://jpa-buddy.com/blog/lombok-and-jpa-what-may-go-wrong/, but can you maybe review this part of the change since you have more experience with it?

goldmann (commented):
I looked at the PR and it looks good, thanks! I just want confirmation from @vibe13 that the `equals` method implementation is what we would expect it to be. Afterwards we can merge it.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/model/RequestEvent.java
vibe13: @dwalluck Since we are changing the equals() to be based on the `id`, I think we should also make the hashcode() rely on the `id` only instead of the hashcode on the whole class?

janinko: ID and hashcode must be consistent over time and persistence state.
Having equals on ID and static hashcode is universal way to achive that.

Since we have IDs always set before persisting to DB, we can also use the ID for hashcode.

vibe13:
@goldmann @dwalluck Apologies I missed the fact that this PR was pending my final confirmation. Yes equals and hashcode methods look good now, we can finally merge this one! Thanks all!

vibe13 (approved):

service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/event/release/ReleaseStandardAdvisoryEventsListener.java
vibe13: I have fixed it as part of https://github.com/project-ncl/sbomer/pull/1164

service/src/main/java/org/jboss/sbomer/service/feature/sbom/errata/event/AdvisoryEventUtils.java
vibe13: Sounds good, thanks!

goldmann (approved):

service/src/main/java/org/jboss/sbomer/service/rest/criteria/predicate/CustomizedPredicateBuilderStrategy.java
vibe13: I need to replace all occurrences of * with %, so yes it could be optimized with `.replace("*", "%")`, but let's note it down and merge this PR.

PR #1137, opened 2025-02-13
fix(SBOMER-337): Allow insecure protocols in Gradle for our cache

vibe13:


dwalluck:
@vibe13 Note that the `allowInsecureProtocol` option was introduced in Gradle 6.0, which isn't a problem if we only run recent Gradle versions.

goldmann (approved):

PR #1138, opened 2025-02-14
[SBOMER-30] - Add SBOM generation for Node.js Yarn builds

vibe13:


goldmann:
Uh, oh, it has a conflict, @vibe13  can you rebase? I'm looking at it now.

goldmann (commented):
I think we may need to update the database after we change the slug for the generator type from `nodejs-cyclonedx` into `npm-cyclonedx`. I agree with this change, btw, it's just that I think we should ensure all the types in the DB refer to the correct type after we switch it.

Thoughts?

vibe13:
> I think we may need to update the database after we change the slug for the generator type from `nodejs-cyclonedx` into `npm-cyclonedx`. I agree with this change, btw, it's just that I think we should ensure all the types in the DB refer to the correct type after we switch it.
> 
> Thoughts?

@goldmann Yes I agree we should polish the DB to be consistent with the name change!

goldmann:
Cool. Other than this, this LGTM. Your intent is to update this PR, right?

vibe13:
@goldmann Yup! Added the SQL script!

goldmann (approved):

PR #1156, opened 2025-03-03
feat(SBOMER-341): add select and search bar for manifests page

thavlice:
I am adding menu and search feature to Manifests page.

This planned to be implemented in the Requests page, but with more filters (there is just one additional in Manifests page).

Notes:

- The options of the select (aka 'No filter' and 'Purl') are represented by string, but probably could be represented by enum/type, which could enable us to add more options easily
- Currently, when no filter is applied, the request string contains `query=` with following value of ` ` (nothing), but this the same effect as not having the `query=` at all.

### DIFF
--- a/ui/src/app/api/DefaultSbomerApi.ts
+++ b/ui/src/app/api/DefaultSbomerApi.ts
@@ -17,7 +17,15 @@
 ///
 
 import axios, { Axios, AxiosError } from 'axios';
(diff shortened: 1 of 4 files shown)


thavlice:
The issues mentioned in the notes previously were fixed.

- No filter now does not add `query=` to the request
- The Select options are represented by enum type.

goldmann (approved):

PR #1163, opened 2025-03-04
feat(SBOMER-81): Add license information to components

dwalluck:
If given an `AnalyzedArtifact`, add any SPDX license identifers to the licenses of the component. Only licenses for which an SPDX license identifier assertion could be made are listed here.

Since URLs are not unique per identifier, do not add any URL to the licenses of the component. Add any valid absolute URLs to the external references of the component.

Add the full list of licenses, including any licenses which have an SPDX license identifier of `NOASSERTION` or `NONE` to the license evidence of the component. The evidence licenses include URL and source information.

### DIFF
--- a/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
+++ b/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
@@ -18,6 +18,7 @@
 package org.jboss.sbomer.core.features.sbom.utils;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.jboss.sbomer.core.features.sbom.Constants.MRRC_URL;
 import static org.jboss.sbomer.core.features.sbom.Constants.PROPERTY_ERRATA_PRODUCT_NAME;
 import static org.jboss.sbomer.core.features.sbom.Constants.PROPERTY_ERRATA_PRODUCT_VARIANT;
 import static org.jboss.sbomer.core.features.sbom.Constants.PROPERTY_ERRATA_PRODUCT_VERSION;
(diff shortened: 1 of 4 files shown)


dwalluck:
It looks like `SbomUtils` was changed a lot unless you use "Hide whitespace" option.

I inverted some `if` statements which reduces the indentation level and I think improves readability.

But, something cleaned up the import of constants. I didn't do that manually.

vibe13 (dismissed):
Thanks @dwalluck this looks good!

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
```diff
@@ -726,10 +826,37 @@ public static void addPedigreeCommit(Component c, String url, String uid) {
...
+        }
+
+        Commit newCommit = new Commit();
+        newCommit.setUid(uid);
+        newCommit.setUrl(url);
+        commits.add(newCommit);
+        pedigree.setCommits(commits);
+
+        c.setPedigree(pedigree);
     }
```
dwalluck, line 856: @vibe13 I tried to rebase to fix merge conflicts. Why is this new code in here? What am I missing?

vibe13: @dwalluck  ? I have no idea, I pushed that code 2 years ago?

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
```diff
@@ -325,8 +334,99 @@ public static void setProductMetadata(Component component, OperationConfig confi
...
+        if (Strings.isEmpty(url)) {
+            return Optional.empty();
+        }
+
+        try {
+            return Optional.of(new URI(url).normalize());
+        } catch (URISyntaxException e) {
+            log.error("Failed to normalize URL '{}': {}", url, e.getMessage(), e);
+            return Optional.empty();
+        }
```
dwalluck, line 347: @vibe13 Just a note: It's possible that the URL in the POM is invalid. Usually, this is due to a case where there's an unexpanded property in the POM. In Build Finder, we don't process every property. We have to "know" which fields to interpolate, but we could interpolate the license URL. I think it should be fixed, but this is outside the scope of this pull request.

https://github.com/project-ncl/build-finder/blob/bf6ed19fc06184d3e6bf0c63df08d419c23a3bb9/core/src/main/java/org/jboss/pnc/build/finder/core/MavenUtils.java#L109-L127

If it fails otherwise, it could be due to a URI that points to a file inside the archive (in which case, I think it's relative, see the check on line 387), which would have to be fixed in Build Finder and not this pull request, so if we start seeing those errors, we can fix it then, too.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
dwalluck: @vibe13 OK, it should be ready to merge after approval, but please see the note I just added about the URLs.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
vibe13: @dwalluck Ah thanks for the details about the url. Sounds good to me, do we need a new release of BF to have the above interpolation working? Thanks. Anyway, beyond this PR, which looks good!

vibe13 (approved):

PR #1164, opened 2025-03-05
[SBOMER-326]: Add advisory identifier to release and build manifests

vibe13:


vibe13:
@goldmann Since this PR is blocked by the broader decision about the taxonomy, should we focus only on the redhat:advisory_id addition here, and then create another PR for the other labels (under discussion)? This way we can unblock the label which was probably required the most.

goldmann:
I think this a good approach, let's go for it!

vibe13:
@goldmann Here it is: https://github.com/project-ncl/sbomer/pull/1197

vibe13:
Let's close this and revisit once we go back to the taxonomy discussion

PR #1172, opened 2025-03-20
[SBOMER-354] - Add container image hashes to "hashes" object

vibe13:


goldmann (approved):
Whoah, the DB change is crazy :)

PR #1173, opened 2025-03-21
Updated build scripts and code formatting

goldmann:


dwalluck:
Something is wrong with the formatter, I think. Why is it not even putting a space after commas?

goldmann:
I know, I asked for ideas why it is happening in the #channel.

dwalluck:
@goldmann I tracked down the cause. You can update to the latest, 2.26.0. You just need to remove the problematic line

```xml
<setting id="org.eclipse.jdt.core.compiler.source" value="1.7"/>
```

from `eclipse-codeStyle.xml`. I tested it, and I think removing it entirely might be better than setting it higher.

goldmann:
Oh, this is great news! I'll update this!

goldmann:
It still changes the output, but I think it's ok.

dwalluck:
> It still changes the output, but I think it's ok.

It did not seem to change the output when I tried all three versions, 2.24.1, 2.25.0, and 2.26.0, `formatter:format` was the same.

It seems like `"1.8"` is what they use in their local copy.

I am not sure how much we differ from their version, but I think I can submit a pull request to add anything missing from our file.

vibe13 (approved):

dwalluck (approved):

PR #1206, opened 2025-04-07
feat(SBOMER-336): update pedigree to use ancestors instead of commits

dwalluck:


dwalluck:
@vibe13 The only issue I have with this is how it is generating the PURL. What about non-GitHub projects?

PR #1208, opened 2025-04-08
Changed cursor behavior to be style 'pointer' only on appropriate elements in UI

thavlice:
Previously the cursor was style 'pointer' (hand pointing) for the each row of the table.
These changes make the cursor style 'auto', meaning we get expected behavior while pointing to certain elements.

These changes could also be implemented via removing the `isClickable` property from `<Tr>` tag, however this wouldn't let us keep a desired feature that highlights the current row on cursor hover.


goldmann (approved):
Nice PR title! 😄

PR #1210, opened 2025-04-09
style: formatter maven plugin

dwalluck:


goldmann:
Does this require code reformat? Or the resulting format is the same and it's just about ensuring we are using "new" settings?

dwalluck:
@goldmann No reformat is required as I have ensured that the latest profile produces the exact same output as before. But, we have seen that any missing settings may have strange defaults and cause unwanted reformatting.

goldmann:
What is the proposal here then? Should we close this one or merge?

goldmann (approved):

PR #1247, opened 2025-04-15
chore: upgrade domino to 0.0.120

aloubyansky:
This release includes a couple of changes for the SBOMer use-case:
* the main metadata component representing a release will depend on every manifested top-level component in an aggregated SBOM, even if the main component doesn't actually depend on those components;
* domino now generates hashes for manifested components, which probably should be disabled for the SBOMer use-case.


goldmann:
I think one more test needs to be fixed:

```
Error:  Failures: 
Error:    SbomerConfigProviderTest.shouldReturnDefaultGenerationConfig:45 expected: <--warn-on-missing-scm --legacy-scm-locator> but was: <--warn-on-missing-scm --legacy-scm-locator --hashes false>
````

goldmann (approved):

PR #1252, opened 2025-04-16
Increase the connection and read timeouts, and add an exponential retry mechanism for clients

vibe13:
Refs: SBOMER-379


goldmann:
What was the reason to go with custom retry implementation and not using SmallRye?

goldmann:
It it about the missing exponential backoff?

vibe13:
@goldmann Yes there is no exponential backoff available

goldmann (approved):

goldmann (approved):
Love it!

PR #1255, opened 2025-04-17
Disable hashes for domino e2e tests

goldmann:
In order to successfully complete e2e tests for Domino we need to disable the new hash calculation feature.


aloubyansky (approved):
Thanks!

PR #1261, opened 2025-04-18
feat(SBOMER-386): Ack UMB messages upon receival and process asynchro…

vibe13:
…nously the requests

### DIFF
--- a/service/src/main/java/org/jboss/sbomer/service/feature/sbom/service/RequestEventRepository.java
+++ b/service/src/main/java/org/jboss/sbomer/service/feature/sbom/service/RequestEventRepository.java
@@ -65,6 +65,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.persistence.Query;
+import jakarta.transaction.Transactional;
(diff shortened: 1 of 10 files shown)


service/src/main/java/org/jboss/sbomer/service/feature/sbom/service/RequestEventRepository.java
```diff
@@ -86,6 +87,14 @@ public RequestEventRepository() {
         super(RequestEvent.class);
     }
 
+    @Transactional
+    public RequestEvent updateWithGenericFailure(String requestEventId) {
+        RequestEvent requestEvent = RequestEvent.findById(requestEventId); // NOSONAR
+        requestEvent.setEventStatus(RequestEventStatus.FAILED);
+        requestEvent.setReason(RequestEvent.FAILED_GENERIC_REASON);
```
goldmann, line 94: Is there a possibility to add more information here? We catch different exceptions which may provide additional context. I think it would be worth to leverage it.

goldmann:
Added only one suggestion, other than this it looks awesome!

goldmann:
I'll merge it now, we can improve in subsequent PR.

goldmann (approved):

PR #1263, opened 2025-04-21
fix: Pyxis empty repository retry

jonathanchristison:
The Pyxis service can return a 200 code but with empty data structure these changes use the smallrye fault tolerance api to retry with a long period fibonacci backoff (days) by using validation on the DTO and throwing a specific exception for the longer lived retries to use

In this commit

* Introduced new Pyxis Wire Mock and tests to create scenarios testing the backoff
* Create a "wrapper"/"delegate" that will call validation after calling the underlying REST Client
* New constants aiming for ~48hr Fn

### DIFF
--- a/service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisService.java
+++ b/service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisService.java
@@ -60,7 +60,7 @@ import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_CLIEN
 @RegisterProvider(PyxisKrb5ClientRequestFilter.class)
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-public interface PyxisClient {
+public interface PyxisService {
 
     List<String> REPOSITORIES_DETAILS_INCLUDES = List.of(
             "data.repositories.registry",
--- a/service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
+++ b/service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
@@ -54,7 +54,7 @@ import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUB
 public class PyxisValidatingClient {
 
     @RestClient
-    PyxisService p;
+    PyxisClient p;
 
     @Inject
     Validator validator;
--- a/service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
+++ b/service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
@@ -45,9 +45,10 @@ import jakarta.enterprise.context.ApplicationScoped;
 
 import jakarta.inject.Inject;
 
-import static org.jboss.sbomer.service.rest.faulttolerance.Costants.PYXIS_UNPUBLISHED_MAX_RETRIES;
-import static org.jboss.sbomer.service.rest.faulttolerance.Costants.PYXIS_UNPUBLISHED_INITIAL_DELAY;
-import org.jboss.sbomer.service.rest.faulttolerance.RetryLogger;
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_MAX_RETRIES;
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_INITIAL_DELAY;
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_MAX_DELAY;
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_MAX_DURATION;
 
 @ApplicationScoped
 public class PyxisValidatingClient {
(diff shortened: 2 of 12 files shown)


service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
```diff
@@ -0,0 +1,100 @@
...
+
+import jakarta.inject.Inject;
+
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_MAX_RETRIES;
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_INITIAL_DELAY;
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_MAX_DELAY;
+import static org.jboss.sbomer.service.rest.faulttolerance.Constants.PYXIS_UNPUBLISHED_MAX_DURATION;
+
+@ApplicationScoped
+public class PyxisValidatingClient {
```
vibe13, line 54: Mmmm `PyxisValidatingService` ? :-)

service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
```diff
@@ -0,0 +1,100 @@
...
+    @BeforeRetry(RetryLogger.class)
+    @FibonacciBackoff(maxDelay = PYXIS_UNPUBLISHED_MAX_DELAY)
+    public PyxisRepositoryDetails getRepositoriesDetails(
+            @PathParam("nvr") String nvr,
+            @QueryParam("include") List<String> includes) {
+        PyxisRepositoryDetails prd = p.getRepositoriesDetails(nvr, includes);
+        Set<ConstraintViolation<PyxisRepositoryDetails>> violations = validator.validate(prd);
+        if (violations.isEmpty()) {
+            return prd;
+        } else {
```
vibe13, line 76: Unnecessary else, you can remove for better readability

service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
```diff
@@ -0,0 +1,100 @@
...
+    @FibonacciBackoff(maxDelay = PYXIS_UNPUBLISHED_MAX_DELAY)
+    public PyxisRepository getRepository(
+            @PathParam("registry") String registry,
+            @PathParam("repository") String repository,
+            @QueryParam("include") List<String> includes) {
+        PyxisRepository pr = p.getRepository(registry, repository, includes);
+        Set<ConstraintViolation<PyxisRepository>> violations = validator.validate(pr);
+        if (violations.isEmpty()) {
+            return pr;
+        } else {
```
vibe13, line 96: same as previous else

service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
```diff
@@ -0,0 +1,100 @@
...
+            maxDuration = PYXIS_UNPUBLISHED_MAX_DURATION,
+            retryOn = ConstraintViolationException.class)
+    @BeforeRetry(RetryLogger.class)
+    @FibonacciBackoff(maxDelay = PYXIS_UNPUBLISHED_MAX_DELAY)
+    public PyxisRepository getRepository(
+            @PathParam("registry") String registry,
+            @PathParam("repository") String repository,
+            @QueryParam("include") List<String> includes) {
+        PyxisRepository pr = p.getRepository(registry, repository, includes);
+        Set<ConstraintViolation<PyxisRepository>> violations = validator.validate(pr);
```
vibe13, line 93: While I can see a `jakarta.validation.constraints.NotEmpty` contraint on the `RepositoryDetail`, I don't see any constraint on the `PyxisRepository`, what condition is supposed to be throwing the exception?

service/src/test/java/org/jboss/sbomer/service/test/PyxisWireMock.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Collections;
+import java.util.Map;
+
+import com.github.tomakehurst.wiremock.WireMockServer;
+
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
+
+public class PyxisWireMock implements QuarkusTestResourceLifecycleManager {
```
vibe13, line 12: I guess you are going to clean this class up? or more changes are expected?

vibe13 (commented):
Hey @jonathanchristison sorry I missed this review pending on me! Apologies for that! All good, some comments picking about names and other minor things, otherwise it looks good! Great work, thanks!!

service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisService.java
```diff
@@ -60,7 +60,7 @@
 @RegisterProvider(PyxisKrb5ClientRequestFilter.class)
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-public interface PyxisClient {
+public interface PyxisService {
```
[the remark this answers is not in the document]
jonathanchristison: Yes, sorry doesn't make a great deal of sense, I was following https://quarkus.io/guides/rest-client#getting-other-response-properties as an example (at the time i was looking at getting the deserialized `RestResponse` response objects but it made it less readable and I didn't really need access to the http response) IMO makes sense to just stick with `PyxisClient` 

I'll do this change as a separate commit so we can always swap it back if two PyxisClients get confusing

service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
vibe13: BTW @jonathanchristison , I just read a comment inside `ReleaseStandardAdvisoryEventsListener` which says that `getRepository` is not used at the moment.... so you can probably nuke it directly :-)

vibe13 (approved):
Great work, thanks Johnny!

PR #1265, opened 2025-04-22
feat(SBOMER-339): include source repo in pedigree

dwalluck:


dwalluck:
@vibe13 My `CachitoClient` implementation is breaking the tests. It requires an existing Build Finder conig and a valid Cachito REST URL. I don't think that "live" tests are allowed, so maybe `CachitoClientIT` should be removed anyway.

PR #1336, opened 2025-05-07
chore(pyxis): Enable rest-client response logging for pyxis

jonathanchristison:
In this commit:
* Changes to `application.yaml` to enable 500 line debug to the pyxis client

Example in logs
```
17:05:55,906 DEBUG traceId= parentId= spanId= [org.jbo.res.rea.cli.log.DefaultClientLogger] (vert.x-eventloop-thread-1) Response: GET http://localhost:36403/v1/images/nvr/jboss-eap-74-openjdk17-builder-openshift-rhel8-container-7.4.20-RELEASED500?include=data.repositories.registry&include=data.repositories.repository&include=data.repositories.tags&include=data.repositories.published, Status[200 OK], Headers[Content-Type=application/json Matched-Stub-Id=490934b1-0a95-4b0d-af09-7febca9855dc Transfer-Encoding=chunked], Body:
                                                                                  {
                            "data": [
                              {
                                "repositories": [
                                  {
                                    "registry": "registry-proxy.engineering.redhat.com",
                                    "repository": "rh-osbs/jboss-eap-7-eap74-openjdk17-openshift-rhel8",
                                    "tags": [
                     mdc:[{}]
```


jonathanchristison:
I think the alternative is to use response filter and call out to logger there if this is a bit too verbose

goldmann:
This is good, but I would expect added logging to the `PyxisValidatingClient` class as well.

service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
```diff
@@ -71,6 +74,11 @@ public PyxisRepositoryDetails getRepositoriesDetails(
         Set<ConstraintViolation<PyxisRepositoryDetails>> violations = validator.validate(prd);
         if (violations.isEmpty())
             return prd;
+        log.warn(
+                "The deserialized Pyxis repositry for nvr: {} and includes:{} violated the following constraints:{}",
```
goldmann, line 78: Nitpicking: a space before second and third  `{}`.

jonathanchristison:
@goldmann not sure if this is what you had in mind? I think Andrea had already covered this with the call to the logger before retry (`@BeforeRetry(RetryLogger.class)`) but I guess this makes it a bit more explicit for the constraint violation, I'd expect either both the retry and the constraint violation to be logged or none as the Retry is _very_ specific.

Example snippet here shows in integration tests we log the retry for network errors (500) and log a retry for the application "error" from the validating client

```
12:11:42,668 WARN  traceId= parentId= spanId= [org.jbo.sbo.ser.res.fau.RetryLogger] (main) Retry attempt for method PyxisClient$$CDIWrapper#getRepositoriesDetails with args [jboss-eap-74-openjdk17-builder-openshift-rhel8-container-7.4.20-RELEASED500, [data.repositories.registry, data.repositories.repository, data.repositories.tags, data.repositories.published]] due to: ClientWebApplicationException: Received: 'Server Error, status code 500' when invoking REST Client method: 'org.jboss.sbomer.service.feature.sbom.pyxis.PyxisClient#getRepositoriesDetails' mdc:[{}]
12:11:42,898 WARN  traceId= parentId= spanId= [org.jbo.sbo.ser.res.fau.RetryLogger] (main) Retry attempt for method PyxisValidatingClient#getRepositoriesDetails with args [jboss-eap-74-openjdk17-builder-openshift-rhel8-container-7.4.20-RELEASED500, [data.repositories.registry, data.repositories.repository, data.repositories.tags, data.repositories.published]] due to: ConstraintViolationException: Pyxis Repository constraint failure mdc:[{}]
```

I think the main thing is the logging response body for later investigation and that _should_ happen via the `PyxisClient` with the changes made to `application.yaml`, it should make it a bit clearer in the logs exactly whats happening here and if we are getting empty data structure from pyxis.

I think an alternative approach is a response filter and put our own logging there and bypass the rest client logging or change the underlying PyxisClient to return RestResponse and handle our validation call and error logging there if either of those approaches make more sense?

service/src/main/java/org/jboss/sbomer/service/feature/sbom/pyxis/PyxisValidatingClient.java
jonathanchristison: This would also annoy me when its time to grep through logs :grin:

goldmann (approved):

PR #1340, opened 2025-05-08
chore: categorize more specific errors upon failures

vibe13:
I was a bit sick of seeing generic messages, now we will log and store more specific errors which will help debugging issues.


goldmann (approved):

goldmann:
Yes, yes, yes!

PR #1381, opened 2025-05-18
New database and API skeleton

goldmann:


service/src/main/java/org/jboss/sbomer/service/leader/LeaderManager.java
```diff
@@ -59,7 +59,7 @@ public LeaderManager(LeaseConfig leaseConfig, KubernetesClient kubernetesClient)
 
     @Scheduled(
             every = "${sbomer.service.leader.check-interval:10s}",
-            delay = 30,
+            delay = 5,
```
goldmann, line 62: Note to self, this should be reverted.

PR #1393, opened 2025-05-22
fix(SBOMER-369): Set expression instead of licenses in evidence when …

dwalluck:
…appropriate


pom.xml
```diff
@@ -124,6 +124,17 @@
...
+        <groupId>org.jboss.pnc</groupId>
+        <artifactId>pnc-api</artifactId>
+        <version>${version.pnc-api}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.jboss.pnc</groupId>
+        <artifactId>pnc-api</artifactId>
+        <version>${version.pnc-api}</version>
+        <classifier>jakarta</classifier>
+      </dependency>
```
goldmann, line 137: I don't see these being used (imported) in the new code.

Second question -- why to have both?

dwalluck: Sorry, the inclusion of both is wrong. It should be the jakarta only. The issue I have is that I need 3.0.7, and neither of the transitive dependencies are providing that. We have it pulled in twice, by core and by cli.

In general, I am not sure why we don't define shared dependencies in `dependencyManagement`, or at least the version properties, in the parent pom. In any case, this is what we currently have:

core (3.0.6 transitively via pnc-common-3.0.1, there is also 3.0.2, still on 3.0.6):

```
+- org.jboss.pnc:pnc-common:jar:jakarta:3.0.1:compile
|  \- org.jboss.pnc:pnc-api:jar:jakarta:3.0.6:compile

```

`org.jboss.pnc:rest-client-jakarta` is also here, but why do I not see `org.jboss.pnc:pnc-api` pulled in?

cli (3.0.7, forced via dependencies):

```
+- org.jboss.pnc:rest-client-jakarta:jar:3.1.5:compile
+- org.jboss.pnc:pnc-api:jar:jakarta:3.0.7:compile
```

goldmann (approved):

PR #1412, opened 2025-06-02
fix: Avoid NPE in release advisory in docker

janinko:
The saveReleaseManifestForDockerGeneration method can be called several times with the same advisoryManifestsRecord which then can lead to updating the same Sbom multiple times. When root purl is regenerated, the first time we can lookup in the originalToRebuiltPurl table the the original root purl of the Sbom, but on subsequent passes, the root purl of the Sbom is already updated, so when we query the originalToRebuilPurl table by the updated root purl, we get null.


vibe13 (approved):
LGTM, thanks!

PR #1484, opened 2025-06-24
fix: Fix PURL code

dwalluck:
* Use `PackageURL::StandardTypes::RPM` constant
* Use `PackageURL::toString` instead of `PackageURL::canonicalize` for consistency
* Look up qualifier keys only once
* Use `PackageURL::toBuilder` insead of constructor for existing PURLs
* Add/fix Syft types: bitnami, terraform, wordpress -> wordpress-plugin
* Replace invalid key character with "-" to match value replacement

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/catalog/SyftImageCatalogCommand.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/catalog/SyftImageCatalogCommand.java
@@ -157,8 +157,11 @@ public class SyftImageCatalogCommand extends AbstractCatalogCommand {
         try {
             PackageURL purl = new PackageURL(component.getPurl());
             Map<String, String> qualifiers = purl.getQualifiers();
-            if (qualifiers != null && qualifiers.containsKey("arch")) {
-                return qualifiers.get("arch");
+            if (qualifiers != null) {
+                String arch = qualifiers.get("arch");
+                if (arch != null) {
+                    return arch;
+                }
             }
         } catch (MalformedPackageURLException e) {
             log.warn("Could not parse the PURL: {}", component.getPurl(), e);
--- a/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/PurlRebuilder.java
+++ b/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/PurlRebuilder.java
@@ -36,6 +36,7 @@ public class PurlRebuilder {
     private static final String SYFT_ALPMPKG = "alpm";
     private static final String SYFT_APKPKG = "apk";
     private static final String SYFT_BINARYPKG = "binary";
+    private static final String SYFT_BITNAMIPKG = "bitnami";
     private static final String SYFT_COCOAPODSPKG = "pod";
     private static final String SYFT_CONANPKG = "conan";
     private static final String SYFT_DARTPUBPKG = "dart-pub";
@@ -49,6 +50,7 @@ public class PurlRebuilder {
     private static final String SYFT_GRAALVMNATIVEIMAGEPKG = "graalvm-native-image";
     private static final String SYFT_HACKAGEPKG = "hackage";
     private static final String SYFT_HEXPKG = "hex";
+    private static final String SYFT_HOMEBREWPKG = "homebrew";
     private static final String SYFT_JAVAPKG = "java-archive";
     private static final String SYFT_JENKINSPACKAGE = "jenkins-plugin";
     private static final String SYFT_KBPKG = "msrc-kb";
@@ -56,7 +58,9 @@ public class PurlRebuilder {
     private static final String SYFT_LINUXKERNELMODULEPKG = "linux-kernel-module";
     private static final String SYFT_NIXPKG = "nix";
     private static final String SYFT_NPMPKG = "npm";
+    private static final String SYFT_OPAMPKG = "opam";
     private static final String SYFT_PHPCOMPOSERPKG = "php-composer";
+    private static final String SYFT_PHPPEARPKG = "php-pear";
     private static final String SYFT_PHPPECLPKG = "php-pecl";
     private static final String SYFT_PORTAGEPKG = "portage";
     private static final String SYFT_PYTHONPKG = "python";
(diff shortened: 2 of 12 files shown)


cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/catalog/SyftImageCatalogCommand.java
```diff
@@ -157,8 +157,11 @@ private String getContainerArch(Component component) {
         try {
             PackageURL purl = new PackageURL(component.getPurl());
             Map<String, String> qualifiers = purl.getQualifiers();
-            if (qualifiers != null && qualifiers.containsKey("arch")) {
-                return qualifiers.get("arch");
+            if (qualifiers != null) {
```
jonathanchristison, line 160: LGTM, added a quick test in additional test in `SyftImageCatalogCommandTest`  while i was trying to figure out whats happening if there's a null arch but probably no need to add to tests 

```
   @Test 
    void missingContainerArch() throws IOException {
        Bom noarch = SbomUtils.fromString(TestResources.asString("boms/catalog/amd64.json"));
        assertNotNull(noarch);
        List<Bom> boms = List.of(noarch);

        String imageName = "rh-osbs/ubi9-ubi-micro";
        String imageDigest = "sha256:1c8483e0fda0e990175eb9855a5f15e0910d2038dd397d9e2b357630f0321e6d";
        
        //Set PURL without arch
        boms.get(0).getComponents().get(0).setPurl("pkg:oci/ubi-micro@sha256%3A213fd2a0116a76eaa274fee20c86eef4dfba9f311784e8fb7d7f5fc38b32f3ef?os=linux&tag=9.4-6.1716471860");

        SyftImageCatalogCommandAlt cataloger = new SyftImageCatalogCommandAlt();
        //We should still be able to progress with just the sbomer:image:labels:architecture property
        assertDoesNotThrow(() -> {cataloger.doCatalog(imageName, imageDigest, boms);});        
        
        //We should throw an application exception if its missing both
        boms.get(0).getComponents().get(0).setProperties(null);
        assertThrows(ApplicationException.class, () -> {cataloger.doCatalog(imageName, imageDigest, boms);});       
    }
    ```

File: core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/UrlUtils.java [hunk 2]
```diff
@@ -43,39 +43,26 @@ public static String urldecode(String value) {
...
      * contain any qualifier that needs to be removed, the original purl is returned, otherwise a new purl is built and
      * returned.
      *
      * @param purl the purl from which the qualifiers should be removed
-     * @param allowList the list of qualifiers which should be removed from the purl
+     * @param removeList the list of qualifiers which should be removed from the purl
      * @return the purl with the qualifiers removed
      */
-    public static String removeAllowedQualifiersFromPurl(String purl, List<String> allowList) {
+    public static String removeAllowedQualifiersFromPurl(String purl, List<String> removeList) {
```
jonathanchristison, line 54: This makes much more sense to me, but some minor nitpicks 

* I don't understand why its `removeAllowedQualifiersFromPurl`  rather than just `removeQualifiersFromPurl` ?

* I think renaming the methods in `UrlUtilsTest.java` for might make tests bit more readable and also helps highlight relevant tests to these changes, they're currently all named "testCaseAllowList" and should probably be "testCaseRemoveList"

jonathanchristison (dismissed):
Thanks @dwalluck these changes make the code more concise, I don't see any issues while stepping through the relevant tests `SbomUtilsTest, `SyftImageCatalogCommandTest`, `SyftImageAdjusterTest`, `UrlUtilsTest` 

I think it would be nice to extend the `shouldSanitizeBogusPurls()` test to catch the bug with for the empty key which you resolved, a purl with a `^=^` qualifier seems to be nicely handled now

I'm setting this to approved as I don't see anything wrong but if you want to address any of the nitpicks then I'll priortise the re-review :+1:  otherwise feel free to merge as is

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/catalog/SyftImageCatalogCommand.java
dwalluck: I added this and credited you in the commit msg.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/UrlUtils.java
dwalluck: I agree. I changed the name.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/PurlRebuilder.java
```diff
@@ -36,6 +36,7 @@ public class PurlRebuilder {
     private static final String SYFT_ALPMPKG = "alpm";
     private static final String SYFT_APKPKG = "apk";
     private static final String SYFT_BINARYPKG = "binary";
+    private static final String SYFT_BITNAMIPKG = "bitnami";
```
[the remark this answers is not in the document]
dwalluck: I added them all. The names are kept the same unless it's listed as a PURL candidate type in which case we use that name.

dwalluck:
> I think it would be nice to extend the `shouldSanitizeBogusPurls()` test to catch the bug with for the empty key which you resolved, a purl with a `^=^` qualifier seems to be nicely handled now

@jonathanchristison Sorry, it's been a while since I created this pull request. Was the empty key "arch" and did it just remove it? I can't recall. I think that packageurl-java will throw an error on a key with an empty value even though according to the spec, I think it's just supposed to ignore it when you call `toString()`. I recall I submitted fixes upstream.

And what did you mean by "^=^` qualifier seems to be nicely handled"?

jonathanchristison:
> > I think it would be nice to extend the `shouldSanitizeBogusPurls()` test to catch the bug with for the empty key which you resolved, a purl with a `^=^` qualifier seems to be nicely handled now
> 
> @jonathanchristison Sorry, it's been a while since I created this pull request. Was the empty key "arch" and did it just remove it? I can't recall. I think that packageurl-java will throw an error on a key with an empty value even though according to the spec, I think it's just supposed to ignore it when you call `toString()`. I recall I submitted fixes upstream.
> 

Yes sorry I let it sit for so long. Ah, I'd assumed there was a deliberate fix was to sanitize both the key and value of an invalid qualifier, previously it only changed the value, it makes sense why the test didn't have this explicitly now.

> And what did you mean by "^=^` qualifier seems to be nicely handled"?
A purl with a qualifier of `^=^` (or any invalid character) previously (before your changes) threw an exception, now it doesn't and it looks to be successfully sanitized.

I think these changes are good to merge (without the addition of testing invalid key and value, though it does handle that now), I've you've got commits for the other minor points thats great, I'll get a re-review done asap once pushed.

PR #1497, opened 2025-07-01
Request Event stuck IN PROGRESS when deliverables used in Notes field for advisories

mertbugrabicak:
Creating a quick draft PR to demonstrate the test that tries to verify things work as it should. Will update with more details when implementation has made more progress


mertbugrabicak:
This is no longer necessary, closing

PR #1503, opened 2025-07-03
Update Text-Only Release SBOM builder to use redhat: prefixed properties

mertbugrabicak:
- Tries to address SBOMER-434
- I updated the test SBOM for deliverables for it to now use redhat: prefixed deliverable-url and deliverable-checksum properties in alignment with the build SBOMs that get produced
- I updated the release sbom building part of the code to now use these redhat: prefixed properties that were also available in Constants


mertbugrabicak:
Thanks for the review and the testing! Merging now and we can try it out tomorrow

PR #1515, opened 2025-07-09
Focus on Pending Request Event upon a DelA Finished

mertbugrabicak:
A fix attempt for SBOMER-426.
This PR:
- Introduces a new test case where a Request Event already triggered for the DelA we were waiting for.
- Tries to solve the issue of a DelA result being connected to 2 request events at once when finished
- Looks if pending requests for this DelA operation exists. If so, prioritizes the request event of the pending request and not the one coming from the UMB. 
- After setting the pending request's request event to be the one to use, ignores the UMB one, citing the reasoning and pointing to the request event that was originally waiting for it.

Basically looks at if it the finished operation has an owner already, if so gives that ownership to them.

### DIFF
--- a/service/src/main/java/org/jboss/sbomer/service/feature/sbom/features/umb/consumer/PncNotificationHandler.java
+++ b/service/src/main/java/org/jboss/sbomer/service/feature/sbom/features/umb/consumer/PncNotificationHandler.java
@@ -248,11 +248,19 @@ public class PncNotificationHandler {
         if (!pendingRequests.isEmpty()) {
             // Get the oldest pending generation request and create a new ConfigMap with the existing id
             pendingRequest = pendingRequests.get(0);
+            // if we have a pending request, use that request event instead and ignore the one from UMB
+            RequestEvent requestEventToIgnore = requestEvent;
+            requestEvent = pendingRequest.getRequest();
+            ignoreRequestEvent(
+                    requestEventToIgnore,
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/jboss/sbomer/service/feature/sbom/features/umb/consumer/PncNotificationHandler.java
```diff
@@ -248,11 +248,21 @@ private void handle(PncDelAnalysisNotificationMessageBody messageBody, RequestEv
         if (!pendingRequests.isEmpty()) {
             // Get the oldest pending generation request and create a new ConfigMap with the existing id
             pendingRequest = pendingRequests.get(0);
+            // if we have a pending request, use that request event instead and ignore the one from UMB
+            RequestEvent requestEventToIgnore = requestEvent;
+            requestEvent = pendingRequest.getRequest();
+            ignoreRequestEvent(
+                    requestEventToIgnore,
+                    "Operation belongs to previous request event " + pendingRequest.getRequest().getId()
```
goldmann, line 256:
```suggestion
                    "Operation belongs to previous request event " + requestEvent.getId()
```

mertbugrabicak: That makes sense, done 👍

mertbugrabicak:
Thanks for the review! Merging now

PR #1533, opened 2025-07-18
SBOMER-431	 Include go standard library packages from Cachi2 remote-sources.json

tecarter94:
https://issues.redhat.com/browse/SBOMER-431


mertbugrabicak (approved):
The CLI part I'm not too familiar with still, but I tried my best to review the code logic, the tests and the resulting SBOM example, and LGTM:
- Well commented to explain what the fields defined are used for
- Nice downloaders and nice tests to cover them
- Nicely makes the go components from the cachi2 depdendencies and adds them to the manifest
- Resulting SBOM looks nice
- Helm part (not here) also LGTM and will approve

PR #1534, opened 2025-07-18
feat: controller and tekton event handler are now handled in a single place

goldmann:


mertbugrabicak:
Great! LGTM after adding the new labels for separation 👍

PR #1537, opened 2025-07-21
Koji Generator for generating RPM manifests in NextGen

mertbugrabicak:
RPM Generator implementation for RPMs based on the SyftGenerator already implemented. 
- Wrote some tests for it to make sure it gets correctly picked up.
- On my local minikube, I'm able to invoke it nicely with:
`{
  "requests": [
    {
      "target": {
        "identifier": "3725358",
        "type": "BREW_RPM"
      }
    }
  ]
}`
where the BREW_RPM type correctly is led to the KojiGenerator type and nicely creates the taskrun for the manifest generation.

Unfortunately, I'm currently having some mac silicon architecture issues with the brew-client image the taskrun uses, so I wasn't yet able to test the manifest generation itself, but the taskrun yaml itself that gets produced looks fine.

Also is dependent on a generators-config.yaml with the generator profile definitions, so will need to wait until that also gets merged

### DIFF
--- a/service/src/main/java/org/jboss/sbomer/service/nextgen/generator/koji/KojiGenerator.java
+++ b/service/src/main/java/org/jboss/sbomer/service/nextgen/generator/koji/KojiGenerator.java
@@ -143,7 +143,7 @@ public class KojiGenerator extends AbstractTektonController {
     @Override
     protected void reconcileGenerating(GenerationRecord generation, Set<TaskRun> relatedTaskRuns) {
         log.debug("Reconcile '{}' for Generation '{}'...", GenerationStatus.GENERATING, generation.id());
-        
+
         boolean inProgress = false;
         boolean success = true;
         TaskRun erroredTaskRun = null;
@@ -341,15 +341,17 @@ public class KojiGenerator extends AbstractTektonController {
 
         TaskRun taskRun = new TaskRunBuilder().withNewMetadata()
                 .withLabels(labels)
+                // TODO: this should be a method
                 .withName(
                         "generation-" + generation.id().toLowerCase() + "-" + SbomGenerationPhase.GENERATE.ordinal()
                                 + "-" + SbomGenerationPhase.GENERATE.name().toLowerCase())
+                .addToAnnotations(ANNOTATION_RETRY_COUNT, "0")
                 .endMetadata()
                 .withNewSpec()
                 .withServiceAccountName(release + SA_SUFFIX)
                 .withTimeout(timeout)
                 .withParams(params)
-                .withTaskRef(new TaskRefBuilder().withName(release + TASK_SUFFIX).build())
+                .withTaskRef(new TaskRefBuilder().withName(release + taskSuffix).build())
                 .withStepOverrides(TektonUtilities.resourceOverrides(request))
                 .withWorkspaces(
                         new WorkspaceBindingBuilder().withSubPath(generation.id())


mertbugrabicak:
I still need to implement the advisory invocation

service/src/main/java/org/jboss/sbomer/service/nextgen/generator/syft/SyftGenerator.java
```diff
@@ -112,12 +113,12 @@ public Set<String> getSupportedTypes() {
 
     @Override
     public String getGeneratorName() {
-        return "syft";
+        return GENERATOR_NAME;
```
tecarter94, line 116: Nice.

service/src/main/java/org/jboss/sbomer/service/nextgen/generator/syft/SyftGenerator.java
```diff
@@ -112,12 +113,12 @@ public Set<String> getSupportedTypes() {
...
     }
 
     @Override
     public String getGeneratorVersion() {
-        return "1.27.1";
+        return GENERATOR_VERSION;
```
tecarter94, line 121: Nice.

service/src/main/java/org/jboss/sbomer/service/nextgen/generator/koji/KojiGenerator.java
```diff
@@ -0,0 +1,391 @@
...
+    public TaskRun desired(GenerationRecord generation) {
+
+        MDCUtils.removeOtelContext();
+
+        log.debug(
+                "Preparing dependent resource for the '{}' phase related to Generation with id '{}'",
+                SbomGenerationPhase.GENERATE,
+                generation.id());
+
+        // TODO: populate traces when we create generations
```
tecarter94, line 296: Not necessary for this PR, but we should really do this at some point in near future.

service/src/main/java/org/jboss/sbomer/service/nextgen/generator/koji/KojiGenerator.java
```diff
@@ -0,0 +1,391 @@
...
+
+        updateStatus(
+                generation.id(),
+                GenerationStatus.FINISHED,
+                GenerationResult.SUCCESS,
+                "Generation finished successfully");
+
+    }
+
+    private TaskRunStepOverride resourceOverrides(GenerationRequest request) {
```
tecarter94, line 265: Bit of a nitpick, but I wonder if we should refactor `resourceOverrides()` to `AbstractTektonController` so that the common method can be called by both `SyftGenerator` and `KojiGenerator`? The methods appear to be identical aside from the use/lack of constants.

service/src/main/java/org/jboss/sbomer/service/nextgen/generator/koji/KojiGenerator.java
```diff
@@ -0,0 +1,391 @@
...
+                                new Quantity(request.generator().config().resources().limits().cpu()),
+                                "memory",
+                                new Quantity(request.generator().config().resources().limits().memory())))
+                .endResources()
+                .build();
+
+    }
+
+    @Override
+    public TaskRun desired(GenerationRecord generation) {
```
tecarter94, line 287: Large chunks of this method are identical to the `desired` method in `SyftGenerator`. It might be good to factor out the identical code into methods in `AbstractTektonController` so that both `desired` methods can call them. Let me know what you think.

tecarter94 (approved):
Overall the changes seem good. But I do think we could de-duplicate quite a bit of code across `KojiGenerator` and `SyftGenerator`.

service/src/main/java/org/jboss/sbomer/service/nextgen/generator/koji/KojiGenerator.java
mertbugrabicak: That makes sense, though since I was mainly using the image generator as a reference to get going, I wasn't sure if it would eventually branch off as the other stuff got added, or if other potential future generators would be similar enough to use a common extracted version of this, but I'll try and figure out what we can move to the abstract controller

mertbugrabicak:
@tecarter94 I have tried to remove some of the duplicate parts.
- Since reconcileGenerating is now shared between the 2 generators and identical, I've made it a concrete shared method in the AbstractTektonController. Possible new generators that differ in approach can still override it.
- I took the stepOverrides and basic label construction out of the generators into a TektonUtilities class.

And lastly, I noticed that the telemetry data being passed was traceId also for the spanId and traceParent variables, I think this might've been a typo, so I made them reflect the spanId and traceParent of the metadata passed instead of traceId for all 3 of them.

- I didn't want to touch the taskrun builder just yet because I don't know how different these can get with future ones.
- I wanted to also share the generate method, but it didn't feel great in the AbstractTektonController mixed in with the controller logic, and it was small enough so I left it as is.

tecarter94:
@mertbugrabicak
> Since reconcileGenerating is now shared between the 2 generators and identical, I've made it a concrete shared method in the AbstractTektonController. Possible new generators that differ in approach can still override it.
Makes sense. Unfortunately you'll have to override `reconcileGenerating` because it does in fact differ for `SyftGenerator` in the main upstream branch. I didn't bother mentioning `reconcileGenerating` in my de-deduplication suggestions due to this reason.
You might want to refactor `GENERATE_OVERRIDE`, `CPU_OVERRIDE`, `MEMORY_OVERRIDE`, `SERVICE_SUFFIX` constants from `SyftGenerator` to `AbstractTektonController` as part of this too, given that they're used in the upstream `SyftGenerator` versions of the methods you refactored.

> I took the stepOverrides and basic label construction out of the generators into a TektonUtilities class.
Nice! Exactly the sort of de-duplication we want.

> And lastly, I noticed that the telemetry data being passed was traceId also for the spanId and traceParent variables, I think this might've been a typo, so I made them reflect the spanId and traceParent of the metadata passed instead of traceId for all 3 of them.
Well spotted! It would have indeed been a typo.

> I didn't want to touch the taskrun builder just yet because I don't know how different these can get with future ones.
Good thinking. It does actually differ already for upstream main's `SyftGenerator` anyway.

> I wanted to also share the generate method, but it didn't feel great in the AbstractTektonController mixed in with the controller logic, and it was small enough so I left it as is.
Fair!

tecarter94:
> @tecarter94 Oh that's a good point! I've put the the reconcileGenerating back to how it was 👍
> 
> I also looked at the GENERATE_OVERRIDE, CPU_OVERRIDE, MEMORY_OVERRIDE, SERVICE_SUFFIX constants you mentioned but I can't really see them in my code base?

@mertbugrabicak See https://github.com/project-ncl/sbomer/blob/fbe695b87ee70836b4ac6a121d00269d6e1224e2/service/src/main/java/org/jboss/sbomer/service/nextgen/generator/syft/SyftGenerator.java#L94

mertbugrabicak:
Oh right! These were some new changes made after I created my PR I think, I had to do a rebase just recently, so I seem to have them now, will take a look 👍

tecarter94:
@mertbugrabicak Just saw the latest commit message. We don't necessarily want the retry mechanism in Kojigenerator, just SyftGenerator. I just thought it might be good to refactor the constants mentioned as they're common concepts across both generators.

mertbugrabicak:
@tecarter94 Sorry for the big new change, when I saw that the syft generator was updated to also have the retry mechanism, I wanted to also add this to the koji generator. While doing it I also tried to remove as much duplicates as I could, moving some under the TektonUtilities and some under the TektonAbstractController.

Since this one went a bit beyond just refactoring, I'll try to run this in my minikube and test it out there to see if it would work as intended, and then report back here

tecarter94:
@mertbugrabicak Sorry to hear you're having difficulty with silicon architecture. I remember having similar problems at my previous job with my M3 Mac!

When testing the retry mechanism locally for `SyftGenerator` I configured it to have a low enough memory request/limit such that it would take 3 retries to successfully complete, so perhaps this is what you could do? It took some fine tuning for the particular container image I was generating for.

1. If the first attempt at running a `TaskRun` fails due to OOM then we enter retry mode.
2. Each time a retry occurs the memory allocated increases by 30%.
3. If 3 retries fail due to OOM then we give up.

Take your time. The changes look good to me, so I'll approve for now. Well done! If worst comes to worst I can eventually test it locally on my Thinkpad with Linux myself.

tecarter94 (dismissed):

tecarter94:
@mertbugrabicak If you rename `SyftRetries` to `Retries` in helm, maybe also consider doing the same rename in [syft-only.yaml](https://github.com/project-ncl/sbomer/blob/main/service/src/test/resources/generator/syft-only.yaml) for consistency?

tecarter94 (approved):

PR #1541, opened 2025-07-22
fix(SBOMER-442): Include checksum for pnc operations

jonathanchristison:
In this commit:

* Added new method in `SbomUtils`
  * `getDistroHashesFromAnalyzedArtifact` takes a AnalyzedArtifact and gets checksums (md5, sha1 & sha256) from AnalyzedDistro
  * A getter AnalyzedDistro to cyclonedx `Hash` mapper is provided
  * This should allow easier additions of other hashing algorithms in future
* Added simple tests for `getDistroHashesFromAnalyzedArtifact`
* During `CycloneDxGenerateOperationCommand` the main component is created and the hashes are set after the creation

### DIFF
--- a/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/CycloneDxGenerateOperationCommand.java
+++ b/cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/CycloneDxGenerateOperationCommand.java
@@ -24,7 +24,7 @@ import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.createBom;
 import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.createComponent;
 import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.createDefaultSbomerMetadata;
 import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.createDependency;
-import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.getDistroHashesFromAnalyzedArtifact;
+import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.getHashesFromAnalyzedDistroribution;
 import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.setArtifactMetadata;
 import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.setPncBuildMetadata;
 import static org.jboss.sbomer.core.features.sbom.utils.SbomUtils.setPncOperationMetadata;
@@ -40,7 +40,6 @@ import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeMap;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
```diff
@@ -412,6 +429,26 @@ private static void addLicenseEvidence(Component component, List<LicenseInfo> li
...
+
+        // Use our pre-defined Algo to getter mapper here (DIST_HASH_DEFINTIONS)
+        for (HashAlgorithmMapping m : DIST_HASH_DEFINITIONS) {
+
+            // eg. Call analyzedArtifact.getDistribution.getSha256 and return a cdx Hash of type SHA_256
+            String hashValue = (String) m.getter.apply(analyzedArtifact.getDistribution());
+
+            if (Objects.nonNull(hashValue) && !hashValue.isEmpty()) {
+                hashes.add(new Hash(m.algorithm, hashValue));
+            }
```
dwalluck, line 447: I don't think that checking `!isEmpty()` is necessary since CycloneDxGenerateOperationCommand.java:174 does not do it and only checks for `null`.

cli/src/main/java/org/jboss/sbomer/cli/feature/sbom/command/CycloneDxGenerateOperationCommand.java
```diff
@@ -171,6 +175,11 @@ protected Path doGenerate() {
                     .map(a -> a.getDistribution().getSha256())
                     .filter(Objects::nonNull)
                     .findFirst();
+
+            distributionHashes = artifactsToManifest.stream()
+                    .map(a -> getDistroHashesFromAnalyzedArtifact(a))
+                    .filter(Objects::nonNull)
+                    .findFirst();
```
dwalluck, line 182: There is no need for `Optional` here, since your method returns an empty list if there are no distribution hashes, which simplifies the call on line 206 to `if (!distributionHashes.isEmpty())) mainComponent.setHashes(distributionHashes)`. I think the issue comes from following the above lines 174--177 too closely. 

I am confused by line 175 (mirrored by line 180) itself. Presumably, there can be multiple artifacts, but they all have the same exact distribution (correct?). Then, why would potentially multiple calls to ` a.getDistribution().getSha256()` or `getDistroHashesFromAnalyzedArtifact(a)` be required? Can't it use something like `distributionHashes = !artifactsToManifest.isEmpty() ? getDistroHashes(artifactstoManifest.get(0).getDistribution()) : Collections.emptyList()`? In any case, you only ever return one list of hashes.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
```diff
@@ -412,6 +429,26 @@ private static void addLicenseEvidence(Component component, List<LicenseInfo> li
         component.setEvidence(evidence);
     }
 
+    public static List<Hash> getHashesFromAnalyzedDistroribution(AnalyzedDistribution analyzedDistribution) {
+        List<Hash> hashes = new ArrayList<>();
+
+        if (analyzedDistribution == null) {
+            return hashes;
+        }
```
dwalluck, line 437: There is a spelling error in the method name here.

Also, you use `Optional.ofNullable` earlier, but this method currently can't return `null`. I think returning an empty list is easier, but if you want to return `null`, then you could move this first `null` check here up to the first line of the method, and also return `null` below if  `hashes` is empty.

dwalluck (approved):

PR #1544, opened 2025-07-23
SBOMER-443 zip component missing from components section

tecarter94:
https://issues.redhat.com/browse/SBOMER-443


mertbugrabicak (dismissed):
LGTM in general 👍 I think it's good that this is consistent across all SBOMs now, and cleaner now that we also wouldn't need to check the deliverable url property for deciding.

I just had one consideration:

From my understanding of it, the first component in the components[] array will have the fully detailed component, while the metadata.component just inherits the type, name and purl. I wonder if this metadata component should also copy the detailed component as closely as possible the other information, such as version, hashes, etc.?

Since the code itself was doing this already for the SBOM types it was duplicating for, even before this change, and it also looks good in Atlas, so I still want to approve. But just something I was thinking about

goldmann:
@tecarter94 I asked Product Security about expectations, but did not get any response so far.

goldmann (approved):

PR #1577, opened 2025-08-01
Carbon UI Draft

thavlice:


ui/src/appV2/components/ManifestsTableTable/ManifestsTable.tsx
```diff
@@ -2,208 +2,128 @@ import { useManifests } from '@appV2/components/ManifestsTableTable/useSboms';
...
+                  <TableRow key={manifest.id}>
+                    <TableCell>
+                      <Link to={`/manifests/${manifest.id}`}>
+                        <pre>{manifest.id}</pre>
+                      </Link>
+                    </TableCell>
+                    <TableCell>
+                      <span>
+                        {manifest?.created
+                          ? timestampToHumanReadable(Date.now() - new Date(manifest.created).getTime(), false, 'ago')
```
jonathanchristison, line 99: I think this is less relevant when events and manifests are happening more frequently but I think the granularity of `timestampToHumanReadable` should be variable. 

`48 days 17 hours 52 minutes ago` is too much information as I'd find the time stamp much more useful, Id suggest 

* If its more than 3 days then only have the days
* If its less than 3 days but more than 1 day have day and hour
* If its less than 1 day have hours and minutes

I think an alternative is to use the ISO datetime string of the creation or trigger date and have the timestamp as a tool tip as its unlikely anyone will need to C&P the human readable delta but its still very useful information
*

goldmann: I like this idea!

goldmann (approved):

PR #1630, opened 2025-08-14
Increase db connections timeouts, lower concurrent generations

mertbugrabicak:
To resolve 503 errors mentioned in SBOMER-469

### DIFF
--- a/service/src/main/resources/application.yaml
+++ b/service/src/main/resources/application.yaml
@@ -154,12 +154,12 @@ quarkus:
   datasource:
     db-kind: postgresql
     jdbc:
-      max-size: 10
+      max-size: 15
       min-size: 0
       initial-size: 1
-      acquisition-timeout: 30
+      acquisition-timeout: 45
       additional-jdbc-properties:
-        connectTimeout: 15
+        connectTimeout: 20
         socketTimeout: 30
 
   # https://quarkus.io/guides/hibernate-orm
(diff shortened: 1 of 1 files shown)


service/src/main/resources/application.yaml
```diff
@@ -154,12 +154,12 @@ quarkus:
   datasource:
     db-kind: postgresql
     jdbc:
-      max-size: 10
+      max-size: 15
```
jonathanchristison, line 157: All these values make sense to me, there is nothing to loose by trying them and we can look to gather more detailed logging/metrics from the datasource in following PRs

service/src/main/resources/application.yaml
```diff
@@ -238,10 +238,10 @@ sbomer:
     generation-scheduler:
       # Defines how many requests can be handled concurrently within the namespace
       # If the capacity is lower than what we have in the DB -- the ones in DB will wait
-      max-concurrent-generations: 20
+      max-concurrent-generations: 15
 
       # Maximum number of generations that will fetched from the database to be scheduled within the namespace.
-      sync-batch: 10
+      sync-batch: 5
```
jonathanchristison, line 244: I think this will definitely help, If we only had 10 connections before and each generation was possibly doing transactional calls (I think quarkus may hold the connection open for this)

jonathanchristison (approved):
LGTM

PR #1631, opened 2025-08-14
Increase connection pool to 20 and decrease concurrent tasks to 10

mertbugrabicak:
In order to attempt to solve the 503 errors


jonathanchristison (approved):
These changes look good to me, I think we probably don't want to exceed the `max-size` of `20` as i'm not sure what would constitute being a bad neighbour to PNC but probably increasing this further :grin:

Lets keep an eye on the db side metrics when we apply this

PR #1727, opened 2025-09-17
SBOMER-452: Remove product version text

jonathanchristison:
Based upon unmerged PR https://github.com/project-ncl/sbomer/pull/1713

* Add additional wiremocks for ET text-only release based on PNC OP
* Change test to verify both erratum wiremocks get a response
* Add additional wiremocks for ET text-only release based on PNC OP
* Change test to verify both erratum wiremocks get a response

Unfortunately this is very difficult to test as even with ET wiremock in dev mode and successful build manifest generation you will hit failures with the Atlas client (even when publish is disabled), I managed to get somewhere with integration tests but hit problems with synthentic data. 

I haven't included this test as It doesn't work and I hit a range of problems related to transactions, it may still be useful as a reference - 

```
package org.jboss.sbomer.service.test.integ.feature.sbom;

import static org.jboss.sbomer.service.feature.sbom.model.RequestEvent.EVENT_KEY_REST_ADDRESS;
import static org.jboss.sbomer.service.feature.sbom.model.RequestEvent.EVENT_KEY_REST_TRACE_ID;
import static org.jboss.sbomer.service.feature.sbom.model.RequestEvent.EVENT_KEY_REST_URI_PATH;
import static org.jboss.sbomer.service.feature.sbom.model.RequestEvent.EVENT_KEY_REST_USERNAME;
import static org.mockito.ArgumentMatchers.anyString;
import static org.mockito.Mockito.mock;
import static org.mockito.Mockito.when;

import java.time.Instant;
import java.util.Map;

import org.jboss.sbomer.core.config.request.ErrataAdvisoryRequestConfig;
import org.jboss.sbomer.core.config.request.RequestConfig;
import org.jboss.sbomer.core.dto.v1beta1.V1Beta1RequestRecord;
import org.jboss.sbomer.core.features.sbom.enums.RequestEventStatus;
import org.jboss.sbomer.core.features.sbom.enums.RequestEventType;
import org.jboss.sbomer.core.features.sbom.utils.ObjectMapperProvider;
import org.jboss.sbomer.service.feature.sbom.model.RequestEvent;
import org.jboss.sbomer.service.feature.sbom.service.AdvisoryService;
import org.jboss.sbomer.service.feature.sbom.service.SbomService;
import org.jboss.sbomer.service.test.ErrataWireMock;
import org.jboss.sbomer.service.test.utils.umb.TestUmbProfile;
import org.junit.jupiter.api.Test;

import com.fasterxml.jackson.databind.node.ObjectNode;

[...]

PR #1728, opened 2025-09-17
fix(SBOMER-452): Remove `product_version_text` requirement

jonathanchristison:
See https://github.com/project-ncl/sbomer/pull/1727 for a more comprehensive PR which includes wiremock payloads but is based on as yet uncommitted changes, this is the minimal set of changes and as of yet its an untested branch of code

In this commit:

* Remove check for and use of `product_version_text`
  * For text only advisories only
  * Identifer replaced with CPE


mertbugrabicak (approved):
LGTM

PR #1742, opened 2025-09-19
SBOMER-480 Fix E2E tests

jonathanchristison:
Previously the E2E tests relied upon the generation endpoint returning a generation, this has been changed to return a request to allow asynchronous calling (we cannot wait for a generation to start before responding to client).

These fixes will result in more calls as we also lookup the generations rather than just using the manifest objects referenced in the request payload, this isn't how we expect the API to be used but resulted in minimal changes to the E2E tests

### DIFF
--- a/e2e/src/test/java/org/jboss/sbomer/test/e2e/E2EBase.java
+++ b/e2e/src/test/java/org/jboss/sbomer/test/e2e/E2EBase.java
@@ -109,13 +109,12 @@ public abstract class E2EBase {
             }
 
             final String status = jsonPath.getString("eventStatus");
-            final String identifier = jsonPath.getString("requestConfig.identifier");
+            final String requestConfig  = jsonPath.getString("requestConfig");
 
             log.info(
-                    "Generation '{}' (type: '{}', identifier: '{}') current status: {}",
+                    "Generation '{}' (requestConfig '{}') current status: {}",
                     requestId,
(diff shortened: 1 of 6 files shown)


mertbugrabicak (dismissed):
LGTM

e2e/src/test/java/org/jboss/sbomer/test/e2e/E2EBase.java
```diff
@@ -167,6 +205,23 @@ public void publishedUmbMessage(String generationRequestId, Consumer<Validatable
...
+        log.info("Fetching request with id '{}'", requestId);
+
+        Response response = RestAssured.given()
+                .baseUri(getSbomerBaseUri())
+                .contentType(ContentType.JSON)
+                .pathParam("id", requestId)
+                .log()
+                .all()
+                .when()
+                .get("/api/v1beta1/requests/id={id}");
```
jonathanchristison, line 218: @dwalluck had a question about this line, it also kinda caught me out as the other endpoints simply take the id without a named parameter required, its also not a query param like we see elsewhere. 

The syntax I think is correct though based off examples in https://stackoverflow.com/questions/32475850/how-to-pass-parameters-to-rest-assured

Also worth pointing out I added raw request logging as I'd hit a few problems (pathParam formatting being one), i figured it would be helpful to leave on

PR #1789, opened 2025-10-15
feat(SBOMER-472): Extract version from generic pkg

jonathanchristison:
Commit 1:

* Add `GenericPurlWrapperUtil`
  * Bit of a mouthfull but acts as a wrapper to Purl adding some convinience methods
  * Attempts to create versioned PURL for generic pkg typ
  * Provides confidence score (how confident we are the regex will be a success) - spoiler very little confidence :-)
  * Provides a difference between two purls
  * Provides a `cyclonedx` `Identity` object
* Changes to `SbomUtils`
  * Add `purlVersionFromGeneric`
    * Takes a component and returns a versioned purl
    * Adds the versioned PURL to the evidence.identities array
    * Gives up at the first sign of trouble (we want to apply this very specifically)
  * Add `appendOrReplaceIdentity`
    * Will either replace purl in evidence.identities (bool replace)
    * Or will append a new evidence.identities

* Tests
  * `GenericPurlWrapperUtilTest` generated unit test class for GenericPurlWrapper
  * `SbomUtilsTest`

Commit 2: 

 Add tests to `SbomUtilsTest`
      * Correct assumptions around where qualifiers end up
      * Test both null purl and null evidences
    * `GenericPurlWrapperUtil`
      * Added a more robust `evidence.identity.method.value` using escaped
        json
    * `GenericPurlWrapperUtilTest`
      * Amended tests to take into account new `method.value` format
    * `SbomUtils`
       * Fixed checks on component purl selection
         * Will choose either purl or evidence.identity based on length
         * If neither exist do nothing
         * If only one exists use that
         * Handle unchanged/unversioned purl

Commit 3:

    * Set/overwrite the evidence PURL for PNC operation derived
      * This will alter the metadata.component
    * Add additional ET wiremocks for local/dev/manual testing

Commit 4:

    * Add versioned purl to evidence identities on release PURL
    * Fix existing tests
      * Fixed assumptions on how many purl identities we should have in the
        case of generic purl

### DIFF
--- a/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
+++ b/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
@@ -48,12 +48,13 @@ import lombok.extern.slf4j.Slf4j;
 public class GenericPurlWrapperUtil {
 
     /*
-     * Help with Jackson Nothing actually uses this yet, the contents of the evidence.identiy.methods.value field are
+     * Help with Jackson Nothing actually uses this yet, the contents of the evidence.identity.methods.value field are
      * not well documented and not a great deal of examples See
      * https://cyclonedx.org/guides/OWASP_CycloneDX-Authoritative-Guide-to-SBOM-en.pdf
      *
      * Taking my best guess here for something useful
      */
+    @Getter
     private static class ValueChanges {
         private final String from;
         private final String to;
@@ -216,8 +217,8 @@ public class GenericPurlWrapperUtil {
 
     /*
      * For generic purls we want to calculate our confidence in the filename method this is our confidence in the
-     * likelyhood of extracting the right version from a given string we apply penalties for things that is likely to
-     * break the regex, such as: number of seperators, variance of seperators, low digit count, high digit count, letter
+     * likelihood of extracting the right version from a given string we apply penalties for things that is likely to
+     * break the regex, such as: number of separators, variance of separators, low digit count, high digit count, letter
      * and digit ratios, these values are tuned by IDEAL_PURLs
      *
      * We will probably never hit 100% confidence as the variation of IDEAL contradict one another
@@ -319,7 +320,7 @@ public class GenericPurlWrapperUtil {
             }
         }
         if (!found) {
-            log.warn("Unable to exctract version from PURL: {}", p.canonicalize());
+            log.warn("Unable to extract version from PURL: {}", p.canonicalize());
         }
         return replaced;
     }


core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
```diff
@@ -0,0 +1,386 @@
...
+public class GenericPurlWrapperUtil {
+
+    /*
+     * Help with Jackson Nothing actually uses this yet, the contents of the evidence.identiy.methods.value field are
+     * not well documented and not a great deal of examples See
+     * https://cyclonedx.org/guides/OWASP_CycloneDX-Authoritative-Guide-to-SBOM-en.pdf
+     *
+     * Taking my best guess here for something useful
+     */
+    private static class ValueChanges {
```
dwalluck: You could use Lombok `@Getter` here although it also looks like nothing is accessing the fields.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
```diff
@@ -0,0 +1,386 @@
...
+    private final double confidenceScore;
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+    /*
+     * Our Regex that covers most the patterns we see on the RCM release area
+     */
+    private static final Pattern[] FILENAME_VERSION_REGEX_STRATEGIES = {
+            Pattern.compile("(?<version>\\d+\\.\\d+\\.\\d+)(?<qualsep>[.-](?<qualifier>Final|[A-Z]+\\d*))?"),
+            Pattern.compile("(?<version>\\d+\\.\\d+)(?<qualsep>[.-](?<qualifier>[A-Z]+\\d*))?"),
+            Pattern.compile("(?<version>\\d+_\\d+)(?<qualsep>[.-](?<qualifier>[A-Z]+\\d*))?") };
```
dwalluck: I see there is a reason for using multiple separate regexes, but maybe you also want to pull out the common parts of the pattern into constants to make it a bit easier to read.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
```diff
@@ -0,0 +1,386 @@
...
+import com.github.packageurl.PackageURL;
+
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class GenericPurlWrapperUtil {
+
+    /*
+     * Help with Jackson Nothing actually uses this yet, the contents of the evidence.identiy.methods.value field are
```
dwalluck: identity

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
```diff
@@ -0,0 +1,386 @@
...
+        idealSeparatorRatio = totalSeparatorRatio / COUNT;
+        idealSeparatorTypes = totalSeparatorTypes / COUNT;
+        idealMaxDigitRun = totalMaxDigitRun / COUNT;
+        idealInterstitialLetters = totalInterstitialLetterCount / COUNT;
+    }
+
+    /*
+     * For generic purls we want to calculate our confidence in the filename method this is our confidence in the
+     * likelyhood of extracting the right version from a given string we apply penalties for things that is likely to
+     * break the regex, such as: number of seperators, variance of seperators, low digit count, high digit count, letter
```
dwalluck: separators (x2)

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
```diff
@@ -0,0 +1,386 @@
...
+                    log.error("Unable to create versioned purl from {}", p.canonicalize(), e);
+
+                }
+
+                found = true;
+                break;
+            }
+        }
+        if (!found) {
+            log.warn("Unable to exctract version from PURL: {}", p.canonicalize());
```
dwalluck: extract

core/src/test/java/org/jboss/sbomer/core/test/unit/GenericPurlWrapperUtilTest.java
```diff
@@ -0,0 +1,214 @@
...
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.packageurl.MalformedPackageURLException;
+import com.github.packageurl.PackageURL;
+
+class GenericPurlWrapperUtilTest {
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    private static class Diff {
```
dwalluck: You could use Lombok `@Getter` here although it also looks like nothing is accessing the fields.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
```diff
@@ -0,0 +1,386 @@
...
+    /*
+     * Values of penalties we can tune
+     */
+    private static final double LENGTH_WEIGHT = 0.005;
+    private static final double DIGIT_RATIO_WEIGHT = 5.0;
+    private static final double LETTER_RATIO_WEIGHT = 1.5;
+    private static final double SEPARATOR_RATIO_WEIGHT = 4.0;
+    private static final double SEPARATOR_TYPES_WEIGHT = 0.1;
+    private static final double MAX_DIGIT_RUN_WEIGHT = 0.2;
+    private static final double INTERSTITIAL_LETTER_WEIGHT = 0.02;
```
dwalluck: TBH, I have no idea if this is good or bad, but I see all the test cases pass. It seems the deliverables don't follow a specific specification for the version, so that doesn't help.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
```diff
@@ -0,0 +1,386 @@
...
+        for (Pattern pattern : FILENAME_VERSION_REGEX_STRATEGIES) {
+            Matcher matcher = pattern.matcher(fileName);
+            if (matcher.find()) {
+                String reconstructedVersion = matcher.group(0);
+                String baseName = fileName.replace(reconstructedVersion, "")
+                        .replace("--", "-")
+                        .replace("..", ".")
+                        .replace("-.", ".")
+                        .replace("__", "_")
+                        .replaceAll("^-|-$", "");
```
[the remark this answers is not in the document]
jonathanchristison: Yea exactly so something like `foo-1.2.3-something.tar.gz` would end up being `foo--something.tar.gz` and lots of other permutations, IIRC the final replaceAll was to handle some files where the version was at the front or the separators bled into the file extension. 

I don't actually have much faith in these regex, there's too many different file naming patterns but we can take a guess and keep a track of what was changed.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/GenericPurlWrapperUtil.java
jonathanchristison: I'm all for more readable but I need an example of what you mean, I think each pattern is subtly different and there's only three, if we add more in the future with more common elements then I think definitely

dwalluck (approved):

Issue #1794, opened 2025-10-24
Duplicate `type` field returned when requesting generations


goldmann:
When we request generation, the returned JSON is not valid, there is a duplicate of the type within `requestConfig` field.

```json
{
    "event": {
        "address": "10.0.0.1",
        "destination": "http://sbomer.domain.com/api/v1beta1/generations",
        "method": "POST",
        "span_id": "8396583bc1f38ae4",
        "trace_id": "db6fd50dcb373c29497aa11d826dcbda",
        "username": "<none>"
    },
    "eventStatus": "IN_PROGRESS",
    "eventType": "REST",
    "id": "F540C0ED8B87411",
    "receivalTime": "2025-10-24T07:23:58.581333036Z",
    "requestConfig": {
        "advisoryId": "155081",
        "apiVersion": "sbomer.jboss.org/v1alpha1",
        "forceBuild": false,
        "type": "errata-advisory",
        "type": "errata-advisory"
    }
}
```

goldmann:
I think the problem is this refactoring: https://github.com/sbomer-project/sbomer/commit/04cd15e938f96fec0f0c4f52f6bb657169a33d12, cc @dwalluck. If we want to keep it like this, I suspect the `include` property here: https://github.com/sbomer-project/sbomer/blob/4a41a9d14471a4131500027de613a3dad7bc73d9/core/src/main/java/org/jboss/sbomer/core/config/request/RequestConfig.java#L40 needs to be adjusted. Other option would be to ignore that field and do not serialize it, as you mentioned.

PR #1796, opened 2025-10-24
fix(#1794): Duplicate type field in requestConfig

jonathanchristison:
This removes the duplicate type field in requestConfig return by excluding it from the json object mapper

### DIFF
--- a/service/src/test/java/org/jboss/sbomer/service/test/integ/rest/RestResourceTest.java
+++ b/service/src/test/java/org/jboss/sbomer/service/test/integ/rest/RestResourceTest.java
@@ -73,7 +73,6 @@ import org.mockito.Mockito;
 
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import io.quarkus.test.InjectMock;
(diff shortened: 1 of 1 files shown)


service/src/test/java/org/jboss/sbomer/service/test/integ/rest/RestResourceTest.java
```diff
@@ -323,6 +327,35 @@ class V1Beta1 {
...
+                        }
+                    } else if (token == JsonToken.END_OBJECT) {
+                        seenKeys.clear();
+                    }
+                }
+                return true;
+            } catch (Exception e) {
+                return false; // Invalid JSON
+            }
+        }
```
dwalluck: I think you can greatly simplify this code:

```java
 private static final JsonFactory JSON_FACTORY = new JsonFactory().enable(JsonParser.Feature.STRICT_DUPLICATE_DETECTION);

        boolean isValidJson(String jsonString) {
            try (JsonParser parser = JSON_FACTORY.createParser(jsonString)) {
                return parser.readValueAsTree() != null;
            } catch (Exception e) {
                return false;
            }
        }
```

and do we want this kind of checking in the class itself as opposed to just the test?

dwalluck: And I changed the name of the method to begin with "is" since it's a `boolean`.

dwalluck (changes requested):

service/src/test/java/org/jboss/sbomer/service/test/integ/rest/RestResourceTest.java
jonathanchristison: Yep much neater and great suggestion, I did have to add objectmapper as the readValueAsTree requires JsonFactory parser be initialized with an ObjectCodec.

service/src/test/java/org/jboss/sbomer/service/test/integ/rest/RestResourceTest.java
```diff
@@ -367,21 +401,23 @@ void shouldRequestPncBuild() {
                     .all(true)
                     .statusCode(202)
                     .extract()
-                    .as(V1Beta1RequestRecord.class);
+                    .response();
 
+            String rawJsonBody = response.asString();
+            V1Beta1RequestRecord record = response.as(V1Beta1RequestRecord.class);
+            assertTrue(validJson(rawJsonBody), "Raw json contains duplicate fields or is invalid");
```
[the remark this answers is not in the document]
jonathanchristison: Yep, avoids unnecessary checks thanks

dwalluck (approved):

PR #1843, opened 2026-04-08
fix(SBOMER-535): sdkman version mismatch

jonathanchristison:
It was found `sdkman install` needs to have an exact match, for some environments we provide only the major version, this on occasion needs to be expanded to a full or partial semver

In this commit
 * `EvironmentAttributesUtils.java`
   * Add static mapping list from maj to semantic
   * Apply the override to `getGradleSDKManCompliantVersion`

### DIFF
--- a/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/EnvironmentAttributesUtils.java
+++ b/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/EnvironmentAttributesUtils.java
@@ -38,9 +38,11 @@ public class EnvironmentAttributesUtils {
      *
      * Add more mappings as needed when we get a mismatch
      */
-    private static final Map<String, String> GRADLE_VERSION_OVERRIDES = Map.ofEntries(
-            Map.entry("9", "9.0.0"), // Highest 9.0.x available in SDKMAN
-            Map.entry("8", "8.0.2") // Highest 8.0.x available in SDKMAN
+    private static final Map<String, String> GRADLE_VERSION_OVERRIDES = Map.of(
+            "9",
+            "9.0.0", // Highest 9.0.x available in SDKMAN
+            "8",
+            "8.0.2" // Highest 8.0.x available in SDKMAN
     );
 
     private static final String SBT_ATTRIBUTE_KEY = "SBT";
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/EnvironmentAttributesUtils.java
```diff
@@ -29,6 +29,20 @@ private EnvironmentAttributesUtils() {
...
+     * (e.g., "9") that need to be expanded to valid SDKMAN versions.
+     *
+     * For single-digit major versions, we map to the highest available X.0.Y patch version for that major version in
+     * SDKMAN (e.g., "9" -> "9.0.0", "8" -> "8.0.2").
+     *
+     * Add more mappings as needed when we get a mismatch
+     */
+    private static final Map<String, String> GRADLE_VERSION_OVERRIDES = Map.ofEntries(
+            Map.entry("9", "9.0.0"), // Highest 9.0.x available in SDKMAN
+            Map.entry("8", "8.0.2") // Highest 8.0.x available in SDKMAN
```
dwalluck: `Map.of()` works too for this few entries.

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/EnvironmentAttributesUtils.java
```diff
@@ -87,11 +101,19 @@ public static Optional<String> getGradleSDKManCompliantVersion(Map<String, Strin
...
+        Optional<String> gradleVersion = environmentAttributes.entrySet()
                 .stream()
                 .filter(entry -> entry.getKey().equalsIgnoreCase(GRADLE_ATTRIBUTE_KEY))
                 .map(Map.Entry::getValue)
                 .findFirst();
+
+        if (gradleVersion.isPresent()) {
+            String version = gradleVersion.get().trim();
+            // Apply version override mapping if available (e.g., "9" -> "9.0.0")
+            return Optional.of(GRADLE_VERSION_OVERRIDES.getOrDefault(version, version));
```
dwalluck, line 115: The first and second args are the same, so what is this meaning to do?

jonathanchristison: for `getOrDefault`? It will return the value matching to key `version` (`9.0.0`) or return value of `version` (`9`)

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/EnvironmentAttributesUtils.java
jonathanchristison: Agreed, changed to Map.of()

dwalluck (approved):

PR #1844, opened 2026-04-10
fix(SBOMER-536): incorrect header and content pair

thavlice:
Switched from using "application/vnd.github.object" as this provided unnecessary metadata and parsing to "application/vnd.github.raw+json", which directly gets the file contents.


jonathanchristison (approved):
Thanks @thavlice LGTM!

PR #1850, opened 2026-06-03
fix(SBOMER-541): Set bomref to be re-adjusted purl

jonathanchristison:
It was found that the bomref diverged from the purl and this causes problems for some tooling that expects these to be identical
     
In this commit:
     * `SbomUtils` update `setPurlVersionFromGeneric` to set bomref for a component where the bomref and previous purl matched

### DIFF
--- a/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
+++ b/core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
@@ -1780,6 +1780,17 @@ public class SbomUtils {
      * @param updateComponentPurl if true, also update the component.purl field (controlled by feature flag)
      */
     public static void setPurlVersionFromGeneric(Component c, boolean updateComponentPurl) {
+        setPurlVersionFromGeneric(null, c, updateComponentPurl);
+    }
+
+    /**
+     * Set the version of a generic PURL by extracting it from the filename.
+     *
+     * @param bom the BOM containing the component (optional, for dependency updates)
+     * @param c the component to update
+     * @param updateComponentPurl if true, also update the component.purl field (controlled by feature flag)
+     */
+    public static void setPurlVersionFromGeneric(Bom bom, Component c, boolean updateComponentPurl) {
         // Grab toplevel and evidence purls
         final Optional<PackageURL> topLevelPurl = Optional.ofNullable(c.getPurl()).map(purl -> {
             try {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
```diff
@@ -1856,11 +1856,17 @@ public static void setPurlVersionFromGeneric(Component c, boolean updateComponen
...
+                            String newPurl = versionedComponentPurl.canonicalize();
+                            c.setPurl(newPurl);
+                            // Also update bomRef to match the new purl, but only if bomRef exists and matches the old
+                            // purl
+                            if (c.getBomRef() != null && oldPurl.equals(c.getBomRef())) {
+                                c.setBomRef(newPurl);
+                                log.debug("Updated component purl and bomRef from {} to {}", oldPurl, newPurl);
+                            } else {
+                                log.debug("Updated component purl from {} to {}", oldPurl, newPurl);
+                            }
```
dwalluck: There's a bug in the `bomRef` comparison and in updating the `bomRef`. The PURL is canonicalized, but not the `bomRef`, and the `bomRef` maybe could be a non-PURL, or at least I would guard against that. The other issue is that if the `bomRef` already exists, e.g., in a dependency, then the change will not propagate.

```java
String newPurl = versionedComponentPurl.canonicalize();
c.setPurl(newPurl);
log.debug("Updated component purl from {} to {}", oldPurl, newPurl);
boolean bomRefMatches = false;
String bomRef  = c.getBomRef();

if (bomRef != null) {
    try {
        bomRefMatches = topLevel.isCanonicalEquals(new PackageURL(bomRef));
    } catch (MalformedPackageURLException e) {
        String oldPurl = c.getPurl();
        bomRefMatches = bomRef.equals(oldPurl);
    }
}

if (bomRefMatches) {
    if (bom != null) { // XXX: need to pass in Bom bom
        SbomUtils.updateBomRef(bom, c, bomRef, newPurl);
    } else {
        c.setBomRef(newPurl);
    }

    log.debug("Updated component purl and bomRef from {} to {}", oldPurl, newPurl);
}
```

dwalluck (changes requested):

core/src/main/java/org/jboss/sbomer/core/features/sbom/utils/SbomUtils.java
dwalluck: Maybe a test like this (I didn't run it):

```java
@Test
void testSetBomRef() {
    String oldPurl = "pkg:generic/jboss-eap-7.4.22-runtime-maven-repository.zip";
    String newPurl = "pkg:generic/jboss-eap-runtime-maven-repository.zip@7.4.22";
    Component c = new Component();
    c.setName("jboss-eap");
    c.setType(Type.FILE);
    c.setPurl(oldPurl);
    c.setBomRef(oldPurl);
    Bom bom = new Bom();
    bom.addComponent(c);
    Dependency d = SbomUtils.createDependency(c.getBomRef());
    bom.addDependency(d);
    SbomUtils.setPurlVersionFromGeneric(bom, c, true);
    assertEquals(newPurl, c.getPurl());
    assertEquals(c.getPurl(), c.getBomRef());
    assertEquals(c.getBomRef(), bom.getDependencies().get(0).getRef());
}
```

jonathanchristison: Thanks @dwalluck some good catches here I hadn't considered, I have added a similar test and reworked the code to use the `updateBomRef` method when the bom parameter is available

dwalluck (approved):

PR #1858, opened 2026-06-22
fix: set skipNotDeployed=false in MavenCycloneDxGenerateCommand

dwalluck:


mertbugrabicak (approved):
LGTM worth testing in stage I think

PR #1861, opened 2026-06-22
fix: fix error message when extracting version from generic PURL

dwalluck:
@jonathanchristison Is `e` preferred here or just `e.getMessage()`? Not the log message right before that uses `e.getMessage()` but we should probably make the text unique in that case to distinguish which one it is.


jonathanchristison (approved):
LGTM

PR #1863, opened 2026-06-24
fix: fix build script parsing

dwalluck:


mertbugrabicak (approved):
LGTM and nicely address all the issues I saw in the examples like quoted arguments, cd inclusions. etc. 

@jonathanchristison Can you also double-check and we can merge and test on stage?

jonathanchristison:
Thanks @dwalluck and @mertbugrabicak at minimum this seems to allow the build to progress further so LGTM!

